PR #58, opened 2023-12-06
custom output path for report.md

TilakMaddy:
Some use cases need the report to be written to a custom location. So I made this change. 
1. It doesn't break existing scripts (since it defaults to "report.md")
2. It generally makes it very convenient for people to integrate in their existing pipelines (maybe you want to write to a file attached as a docker volume, etc) 

```bash
Rust based Solidity AST analyzer

Usage: aderyn [OPTIONS] <ROOT>

Arguments:
  <ROOT>  Foundry or Hardhat project root directory

Options:
  -o, --output <OUTPUT>  Desired file path for the final report (will overwrite existing one) [default: report.md]
  -h, --help             Print help
  -V, --version          Print version
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /target
+.DS_Store
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,7 +16,7 @@ use crate::detect::detector::{get_all_detectors, IssueSeverity};
 use crate::report::printer::{MarkdownReportPrinter, ReportPrinter};
 use crate::report::reporter::{Issue, Report};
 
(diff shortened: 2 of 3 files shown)


TilakMaddy:
More importantly this help you maintain different versions of your output. 

So the workflow is 

1. I run `aderyn . -o v1.md`
2. I see what changes need to made 
3. Then I run `aderyn -o v2.md`
4. so on ...

I'm also open to auto naming the output names with timestamp (like how foundry does for dry runs) and include the `latest.md` or someth like that

alexroan (approved):

PR #65, opened 2023-12-08
python bindings

TilakMaddy:
With this PR you can use `maturin` [http://maturin.rs](http://maturin.rs) to build python packages for any OS architecture (linux, mac & windows) then distribute it. (typically created  in  workflow config of CI/CD)

Users would only need foundry/hardhat installed and python. (No cargo needed)

Then they can do this:

```python
import aderynpy

def analyze_vulnerabilities_for_file():
    aderynpy.generate_report("/path/to/SimpleStorage.sol", "analysis.md")

def analyze_vulnerabilities_for_project(root, output):
    aderynpy.generate_report(root, output)
```


TilakMaddy:
Clarification - I have introduced another layer of workspace abstraction to support these different bindings

So we have :
                                  -------> aderyn (cli)
                                  |
core --->  driver ---- -----> aderyn (py bindings)
                                 |
                                 -------> aderyn (node js bindings)

alexroan:
This looks great.

Up to this point, I've been manually releasing the crates locally (`cargo publish`). Given that `maturin` requires a separate release flow, the next step after this should probably be to create a formal release flow for both the crates and the python bindings.

What are your thoughts on how that should be setup?

alexroan (approved):

PR #71, opened 2023-12-09
init prototype (not intended for merge)

TilakMaddy:
Hey everyone! I was wondering if this idea aligns with the project's interests. ....

So basically I was thinking to myself what if users can write their own detectors and leverage the processing power of aderyn to generate their reports without having to rely on the "official" detector modules. 

So to get a small taste of that would look like, I have created a repo that I would like for you guys to try out - set it up locally and let me know how you feel ....

Link to the repo:

(https://github.com/TilakMaddy/custom_detector_example)

[https://github.com/TilakMaddy/custom_detector_example]

You only have to change the `root_path` in `main.rs` to point to some local foundry project and do cargo run.


alexroan:
This is a great initiative.

It reminds me of `solhint` rules in JS, where you can define your own custom listing rules, but also inherit the official rules. Does this also do both? Or does it only allow mutually exclusive sets?

TilakMaddy:
> This is a great initiative.
> 
> It reminds me of `solhint` rules in JS, where you can define your own custom listing rules, but also inherit the official rules. Does this also do both? Or does it only allow mutually exclusive sets?

Thanks for asking ! 
So the current example that I have shown runs only the custom detectors (that you add in `main.rs` ) but that doesn't mean it can't be extended to the official detector modules as well. You can completely subscribe to what you want !

TilakMaddy:
> > This is a great initiative.
> > It reminds me of `solhint` rules in JS, where you can define your own custom listing rules, but also inherit the official rules. Does this also do both? Or does it only allow mutually exclusive sets?
> 
> Thanks for asking ! So the current example that I have shown runs only the custom detectors (that you add in `main.rs` ) but that doesn't mean it can't be extended to support the official detector modules. You can completely subscribe to what you want !

We would then also have to think about tracking versions of each detector module independently

alexroan:
Hmmm, OK. I don't want to add too much project complexity just yet, versions for each detector might be overkill right now.

TilakMaddy:
Tracking new version here https://github.com/Cyfrin/aderyn/pull/123

PR #72, opened 2023-12-10
option for json output available

TilakMaddy:
With this PR, if your output file name ends with `.json` instead of `.md` you will get a json file output. 

(everything else still holds Eg: default is still report.md, *.md will give .md files, etc, )

I did this because I wanted to help write tests in foundry like shown here 
[https://github.com/Cyfrin/foundry-devops/pull/13](https://github.com/Cyfrin/foundry-devops/pull/13)


aderyn_core/src/report/markdown_printer.rs
```diff
@@ -0,0 +1,262 @@
...
+
+use super::{
+    printer::ReportPrinter,
+    reporter::{Issue, Report},
+};
+
+pub struct MarkdownReportPrinter;
+
+impl ReportPrinter<()> for MarkdownReportPrinter {
+    fn print_report<W: Write>(
```
alexroan, line 12: Could we print each of the issues in the JSON file, also?

As well as `issues_found` summary section, another section called `issues` that has members: `critical`, `high`,... etc. Then within each of those, an array of every issue found.

alexroan (changes requested):

aderyn_core/src/report/markdown_printer.rs
alexroan: In fact, it makes sense to produce a JSON output that mimicks the MD:
```json
{
  "files_summary": {...},
  "files_details": {...},
  "issue_summary": {...},
  "critical_issues": {...},
  "high_issues": {...},
  ...
}
```

TilakMaddy: Done 👊

TilakMaddy: Oh wait..there's 1 more change i wanna make

TilakMaddy: Done 👊

aderyn_core/src/report/markdown_printer.rs
TilakMaddy, line 1: hey @alexroan don't pay too much attention to this file - I have changed it quite a bit in the next PR (where we make cmd+click links) - that is heavily refactored

PR #73, opened 2023-12-10
Feature/ctrl click visit file

TilakMaddy:
.

it's ready to go but will treat it as draft until https://github.com/Cyfrin/aderyn/pull/72 is approved. (because it relies on some of the newly created abstractions)


TilakMaddy:
update ! -> i just got done adding line preview section too : )

TilakMaddy:
image

TilakMaddy:
Ready to go ! (after json output)

aderyn_core/src/report/markdown_printer.rs
```diff
@@ -0,0 +1,222 @@
...
+        {
+            for (index, issue) in issues.iter().enumerate() {
+                let issue_title_slug = issue
+                    .title
+                    .to_lowercase()
+                    .replace(' ', "-")
+                    .replace(|c: char| !c.is_ascii_alphanumeric() && c != '-', "");
+                writeln!(
+                    w,
+                    "  - [C-{}: {}](#C-{}-{})",
```
alexroan, line 135: This currently prints `C` for all severities.

aderyn_core/src/report/markdown_printer.rs
```diff
@@ -0,0 +1,222 @@
...
+            let line_preview = line
+                .split('\n')
+                .skip(instance.line_no - 1)
+                .take(1)
+                .next()
+                .unwrap();
+
+            writeln!(
+                writer,
+                "- Found in {} Line: {}\n\n\t>{}\n\nfile://{}\n\n\n",
```
alexroan, line 215: I'm wondering if we make the line preview use a code snippet format instead of a quote format.

So instead of:

>            target.delegatecall(abi.encodeWithSignature("doSomething(uint256)", I));

We print:
```solidity
            target.delegatecall(abi.encodeWithSignature("doSomething(uint256)", I));
```

aderyn_core/src/report/markdown_printer.rs
```diff
@@ -0,0 +1,222 @@
...
+                instance.contract_path, instance.line_no, line_preview, &path,
```
alexroan, line 216: Screenshot 2023-12-11 at 11 08 55


I think we're going to have to print the entire file path out here. The edge cases are that aderyn is run on a directory not currently loaded in VSCode, or the relative path doesn't resolve to anything.

alexroan (changes requested):
At first glance, just a few small changes

aderyn_core/src/report/markdown_printer.rs
TilakMaddy: Yes, that seems to bring out the correct semantic meaning. You are right!

aderyn_core/src/report/markdown_printer.rs
TilakMaddy: Thanks for the catch! ✅

TilakMaddy: just some backstory - I use `~` in my mac terminal and zsh resolves that to the full path... so the actual program wouldn't know about it - for the program, it thinks it has been given the absolute full path.  I should've been more careful :P

aderyn_core/src/report/markdown_printer.rs
```diff
@@ -0,0 +1,231 @@
...
+            let line_preview = line
+                .split('\n')
+                .skip(instance.line_no - 1)
+                .take(1)
+                .next()
+                .unwrap();
+
+            writeln!(
+                writer,
+                "- Found in {} Line: {}\n\n\t```solidity\n\t{}\n\t```\nfile://{}\n\n\n\n",
```
alexroan, line 221: Could we append the line number in the `file://` so that on ctrl+click it opens the file on the correct line?

alexroan (commented):
Looking at the new report.md, I like the ctrl+click, but I'm worried that the line_preview bloats the file and doesn't add much value for the user.

@TilakMaddy What do you think about removing the line_preview, keeping the ctrl+click, appending the line number to the file:// link so that it opens the correct line, and using the "Found in Line: ..." to insert a hyperlink?

aderyn_core/src/report/markdown_printer.rs
alexroan: Also, if we remove the line_preview, we should make the `file://` url a markdown link. As in:

```md
- Found in [Line: 35](file://path/to/file.sol:35)
```

TilakMaddy: Hi Alex! I have changed it a bit. Please can you check again and see if you like it? The links are now clickable from within the markdown-rendered view.  However, I didn't remove the line preview yet : ) 

So now, that we're not printing `file:///blah/blah/blah` it conserves some space. And if you still think line preview isn't going to add too much value I'll remove it

aderyn_core/src/report/markdown_printer.rs
TilakMaddy: @alexroan - I believe not having lowercase is why report tests are failing.

aderyn_core/src/report/markdown_printer.rs
```diff
@@ -218,11 +229,17 @@ impl MarkdownReportPrinter {
 
             writeln!(
                 writer,
-                "- Found in {} Line: {}\n\n\t```solidity\n\t{}\n\t```\nfile://{}\n\n\n\n",
+                "- Found in {} [Line: {}]({}#L{})\n\n\t```solidity\n\t{}\n\t```\n",
```
TilakMaddy, line 232: `#L` here stands for line number. A colon followed by number doesn't work here. 

tldr;
The link is sent to the vs-code server running remotely to resolve to the file. And you can imagine in a URL you add # to indicate sort of an ID. When you use colon, it breaks URL (firstly because it's not URLencoded and 2nd - it implies you are trying to connect to a port of some sort)..That's why you get HostNotFound type of errors

alexroan: Makes sense, thank you for the explanation

aderyn_core/src/report/markdown_printer.rs
alexroan: This looks great!

alexroan:
@TilakMaddy You'll notice a new workflow still failing. To resolve this, run:

`cargo run -- ./tests/contract-playground/`

and

`cargo run -- -o report.json ./tests/contract-playground/`

and push the results.

This is so that we can see, in the Git diff, how changes to the code affect the report. :)

TilakMaddy:
Oops looks like I messed up in resolving conflicts :P - I just copy pasted the report.md and report.json from local file into github website..

TilakMaddy:
Contd here - https://github.com/Cyfrin/aderyn/pull/80

alexroan:
Merged #80 instead of this

PR #85, opened 2023-12-13
Detector: Arbitrary `from` passed to `transferFrom`

alexroan:
* Arbitrary `from` passed to `transferFrom`
* Fix overzealous solmate detector


report.md
```diff
@@ -126,6 +127,42 @@ https://github.com/transmissions11/solmate/blob/main/src/utils/SafeTransferLib.s
 `@dev Note that none of the functions in this library check that a token has code at all! That responsibility is delegated to the caller`
 
 
+- Found in src/ArbitraryTransferFrom.sol [Line: 20](tests/contract-playground/src/ArbitraryTransferFrom.sol#L20)
```
alexroan, line 130: These shouldn't be caught by the solmate detector.

report.md
```diff
@@ -423,6 +478,12 @@ Solc compiler version 0.8.20 switches the default target EVM version to Shanghai
 	pragma solidity 0.8.20;
 	```
 
+- Found in src/ArbitraryTransferFrom.sol [Line: 2](tests/contract-playground/src/ArbitraryTransferFrom.sol#L2)
```
alexroan, line 481: This shouldn't be caught by the PUSH0 detector

report.md
alexroan: This happens because the somate detector first checks if there are ANY solmate imports across the whole project, then finds all references to:

```
if member_access.member_name == "safeTransfer"
                || member_access.member_name == "safeTransferFrom"
                || member_access.member_name == "safeApprove"
```

Solutions:
- Check for solmate imports and recommend against -> This would alter the solmate detector to find import statements, and not the actual references -> some false positives
- Check that the references to these functions only occur in source_units where the solmate dependency is imported -> some false negatives (worse)

report.md
alexroan: Tracking here: #88

Issue #34, opened 2023-11-11
nSLOC is not correct


PatrickAlphaC:
```
git clone https://github.com/Cyfrin/6-thunder-loan-audit
cd 6-thunder-loan-audit
make 
aderyn .
```

You'll see the `nSLOC` output is incorrect

alexroan:
Could you be more specific? What is the expected value, and what output do you get?

PatrickAlphaC:
The output I get is as follows for nSLOC.

```
# Summary

## Files Summary

| Key | Value |
| --- | --- |
| .sol Files | 8 |
| Total nSLOC | 103 |
```

However, this is the correct nSLOC values:

<img width="1477" alt="Screenshot 2023-11-19 at 9 06 57 PM" src="https://github.com/Cyfrin/aderyn/assets/54278053/a10871f7-89be-4e4a-9796-b6e70db434fa">

alexroan:
Could you give access to that repo?

alexroan:
Interesting, the `tokei` package that we use to get details about the source only reads down to line 23 in `src/protocol/ThunderLoan.sol` because of this comment syntax:

```
    /*//////////////////////////////////////////////////////////////
                            STATE VARIABLES
    //////////////////////////////////////////////////////////////*/
```

it's struggling with `/*////////...`

alexroan:
Issue submitted to tokei [here](https://github.com/XAMPPRocky/tokei/issues/1050)

alexroan:
Fixed by #93 ! 🥳

PR #93, opened 2023-12-16
correctly calculate nSLOC

TilakMaddy:
Proposing fix for https://github.com/Cyfrin/aderyn/issues/34

`tokei` tries to support a lot of languages and myriad of comment styles. So as a result when you try to make changes to the codebase for desiring the behavior I have described [here](https://github.com/XAMPPRocky/tokei/issues/1050#issuecomment-1853524571)  a lot of other things start to break. (I'm sure the core maintainers of tokei are much smarter than I am - here I am simply talking about experimenting with surface level changes to tokei)

Hence I decided to create my own parser (added in this PR) . Also, I have included 4 test solidity files one of which is the Thunder Loan contract - and I have manually verified the lines of code in each one of them.

Hope this helps!


Cargo.lock
TilakMaddy, line 1: 30% reduction in release binary size after removing tokei!  Look inside this to see what we were able to get rid of : )

tests/solfiles/program2.sol
```diff
@@ -0,0 +1,77 @@
...
+     this is longer comment */ uint256 s_5 = 0;
+
+    function foo(address bar) public pure {
+        // blah ouhsff 
+        // sdfssaf /*dsf
+
+        require(bar != address(0));
+    }
+
+    function emoji() public pure {
```
TilakMaddy, line 38: lifted this example from solidity docs

tests/solfiles/program2.sol
```diff
@@ -0,0 +1,77 @@
...
+        OOBLAH
+    /////*/
+        string memory simple3 = '/* this is not a \\\' simple comment */'; // this semicolon no longer disappears !!
+        // h haha sd
+        string memory simple4 = '/* this is not \' /*a comment */';/* this is a comment */ uint256 m_7 = 0;
+         /*//////// 
+        ANOTHER ONE
+    /////   */
+        require(bar != address(0));
+
```
TilakMaddy, line 57: this way of writing strings is also something I have seen in the docs

alexroan (commented):
The numbers look good! Remaining TODOs:
* One review comment to resolve on test file location
* Some conflicts to resolve before merging.

Other than those two, this looks great!

tests/solfiles/program.sol
```diff
@@ -0,0 +1,60 @@
+/*
+    normal comment
```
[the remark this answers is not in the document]
TilakMaddy: Not a big reason - It's just that the Thunderloan contract import a few external libraries and it felt unnecessary to bring in those in as submodules and compile have them compile when all we're trying to do is count lines of code.

But in retrospect, I think we will have stronger guarantees that the code is valid (even though it's not 100% prerequisite) if we have them inside the `./tests/contract-playground/src`. After all we expect that forge build works !

aderyn_core/src/report/mod.rs
```diff
@@ -109,7 +109,7 @@ impl ContextLoader {
     pub fn files_summary(&self) -> FilesSummary {
         FilesSummary {
             total_source_units: self.source_units.len(),
```
TilakMaddy, line 111: @alexroan  please can you clarify that `source_unitls.len()` will be equal to `src_filepaths.len()` that you newly added ?

TilakMaddy: If yes, then I think that we need to somehow enforce that equality in code by making another abstraction somewhere coz I am generally a fan of single source of truth... But I also understand that it's not always feasible :P

tests/solfiles/program.sol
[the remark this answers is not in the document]
TilakMaddy: I am contemplating not having thunder loan... its too heavy to integrate and it's quite redundant given the other 3 test files have quite complex cases covered. I'll add those 3 to the contract playground

aderyn_core/src/report/mod.rs
alexroan: Great question...

I learnt yesterday that `source_units.len()` != `src_filepaths.len()`.

Tl;dr:
* `source_units.len()` = the total number of AST source units loaded (there may be duplicates based on solc version)
* `src_filepaths.len()` = the total number of solidity files.

Why?:

When Foundry compiles the source files, sometimes Foundry compiles and outputs to multiple files, like this:
image

This occurs when more than one specific solidity pragma version is defined in contracts, and there are also contracts with wide pragmas.

Aderyn loads both ASTs, even though they represent the same source file. We _have_ to load both because of the way the AST objects within them reference each other.

TilakMaddy: ✌️

tests/solfiles/program.sol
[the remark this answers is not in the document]
TilakMaddy: Done!

report.md
```diff
@@ -632,6 +652,36 @@ Assigning values to address state variables without checking for `address(0)`.
 	    function setNonEmptyAlteredNumbers(
 	```
 
+- Found in src/cloc/AnotherHeavilyCommentedContract.sol [Line: 31](tests/contract-playground/src/cloc/AnotherHeavilyCommentedContract.sol#L31)
+
+	```solidity
+	    function foo(address bar) public pure {
```
alexroan, line 658: Note for me: This is a false positive that should be fixed.

alexroan (commented):
Hmmm, I'm running this branch on some other codebases and I'm getting this:

```
## Files Summary

| Key | Value |
| --- | --- |
| .sol Files | 0 |
| Total nSLOC | 14754 |
```

Potentially a bad conflict resolution from `dev` @TilakMaddy ?

alexroan (approved):

PR #101, opened 2023-12-21
introducing ContextBrowser as an abstraction on ContextLoader for detectors 

TilakMaddy:
```rust

         // Get all function definitions
        for function_definition in loader.function_definitions.keys() {
            // Get all the binary checks inside the function
            let binary_checks: BinaryChecks = function_definition.into();
            // {...}
            // Get all the assignments in the function
            let assigments: Assignments = function_definition.into();
            // {...}
        }

```

This PR will allow us to climb up this mountain
image

* ref: [https://craftinginterpreters.com/a-map-of-the-territory.html](https://craftinginterpreters.com/a-map-of-the-territory.html)

So far we were in step 5 which is the syntax tree (`ContextLoader`) - however for detectors to analyze the code comfortably we will need to take it up a notch to analysis (which is the peak) after which you go down again ...

That peak in the mountain is what I would like to  represent `ContextBrowser` with - it is passed along as a parameter to the `detect` function and it seeks to help make more friendly queries.

In this PR I have integrated that in the file `zero_address_check.rs` which was supposedly "Crazy Complex" Hopefully it is not that now !

If this works out, we can slowly start to integrate it for other detection modules : )

### DIFF
--- a/aderyn_core/src/context/browser/assignments.rs
+++ b/aderyn_core/src/context/browser/assignments.rs
@@ -1,0 +1,29 @@
+use eyre::Result;
+
+use crate::{
+    ast::{Assignment, FunctionDefinition},
+    visitor::ast_visitor::{ASTConstVisitor, Node},
+};
+
+pub struct Assignments {
+    pub assignments: Vec<Assignment>,
+}
+
+impl From<&FunctionDefinition> for Assignments {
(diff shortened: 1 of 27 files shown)


alexroan (approved):
Another epic feature from @TilakMaddy <3

PR #102, opened 2023-12-22
Detector: uninitialized state variables

giovannidisiena:
WIP attempt at adding an uninitialized state variables detector.

@alexroan this is what I have so far. It's currently not finding any instances, I believe due to the implementation of `UninitializedStateVariablesDetector::function_uses_variable_without_initialization` as it was finding every instance previously before adding this logic.

I also noticed that you implement `ASTConstVisitor` for `ZeroAddressCheckDetector` - what was the motivation for that and do I need to do something similar?


aderyn_core/src/detect/high/uninitialized_state_variables.rs
```diff
@@ -0,0 +1,156 @@
...
+            }
+        }
+
+        // If no uninitialized usage is found, return false.
+        false
+    }
+}
+
+impl Detector for UninitializedStateVariablesDetector {
+    fn detect(&mut self, loader: &ContextLoader) -> Result<bool, Box<dyn Error>> {
```
alexroan, line 68: Good start, but there's probably an easier way to approach this one. Instead of starting with the assignments, what if it:

1. Get all state variables that are not hardcoded with a value - into a variable called `found_state_variables`
2. Get all Assignments - `found_assignments`
3. For each of the Assignments, if the left-hand side references a `found_state_variable`, remove it from the `found_state_variables`.

Whatever is left inside `found_state_variables, add them to the `instances`. W probably don't need the extra `impl` above, and we can add specifically the state variables declaration to the `instances`.

Make sense?

giovannidisiena: So I decided to check all function bodies independently as an assignment in one would mean that uninitialized usage in another would not be considered a found instance. Are you suggesting that perhaps this is desired, e.g. admin setter function exists, therefore usage in other functions should not be uninitialized and we should not detect this as an issue?

alexroan: Yep, once we have that working, we can discuss what it finds and whether or not we should rethink how granular it should be

alexroan:
Closing this in favour of #103

PR #103, opened 2023-12-22
Detector: uninitialized state variables (simplified)

giovannidisiena:
@alexroan here is the simplified version of this detector. Though I am still running into a couple of issues:
1. We need to find a way of determining whether the state variable is ever used, but I am not sure how best to achieve this. Currently, we have uninitialized state variables that are never used contribute to the found instances count (test is currently failing because it's finding 4 instances but we should only have 1 uninitialized state variable being used).
2. We probably also want to modify the detector to insert the instances where the uninitialized state variable is actually used, since the count will be off (e.g. we should actually have 2 instances of uninitialized state variable `addr2` being used in both `useUninitializedAddressForTransfer` and `useUninitializedAddressForCall`).

Does that make sense to you?


alexroan:
This will need to be rebased from `dev` @giovannidisiena

alexroan:
Closing due to staleness @giovannidisiena . We can pick up at a later date once the core architecture is a little more refined.

PR #119, opened 2024-01-05
Fix/line number issues

### LINKED ISSUE #118: Bug: Non-ASCII char in comments screw up line number searching and resulting snippet generation
The solc compiler output generated positions that are based on single byte level (I think).

The way aderyn reads it is different, resulting in wrong line numbers being extracted from the source, reporting line numbers and snippets that don't represent the actual issue.

alexroan:
Fixes #118


aderyn_core/src/context/browser/node_locator.rs
```diff
@@ -1,51 +0,0 @@
-use super::{ContextBrowser, SourceUnitInsight};
```
alexroan, line 1: @TilakMaddy I walked back the node_locator here in favour of the old method, because it was causing incorrect line numbers and file snippets to be produced with non-ascii chars (see #118 ).

It also slims down the dependency injection into detectors, so that the extra `ContextLoader` doesn't need to be passed in yet, but we can still "step-into" child nodes using `.into()`.

report.md
```diff
@@ -754,10 +826,10 @@ Assigning values to address state variables without checking for `address(0)`.
 
 Index event fields make the field more quickly accessible to off-chain tools that parse events. However, note that each index field costs extra gas during emission, so it's not necessarily best to index the maximum allowed per event (three fields). Each event should use three indexed fields if there are three or more fields, and gas usage is not particularly of concern for the events in question. If there are fewer than three fields, all of the fields should be indexed.
 
-- Found in src/eth2/DepositContract.sol [Line: 50](tests/contract-playground/src/eth2/DepositContract.sol#L50)
+- Found in src/eth2/DepositContract.sol [Line: 19](tests/contract-playground/src/eth2/DepositContract.sol#L19)
 
 	```solidity
-	    ///  `interfaceId` is not 0xffffffff, `false` otherwise
+	    event DepositEvent(bytes pubkey, bytes withdrawal_credentials, bytes amount, bytes signature, bytes index);
```
alexroan, line 832: @TilakMaddy an example of incorrect snippet due to incorrect line number.

aderyn_core/src/context/browser/node_locator.rs
TilakMaddy: @alexroan First of all, I'm sorry, I think that it was a bad decision on my side and a bit of carelessness :P  (I need to investigate the line number thing further but good that you removed it - 'twas a micro optimization; I also feel it's not needed)

I can totally see how we can use the context browser's functionality without having to inject it and I think it's great.
My only concern would be if any "post-processing" of the AST would be required  to do things like you mentioned "stepping out"  to the parent nodes. 

Well one solution could be, to use the "loader" injected as a dependency to do the post-processing - but,  the problem if there is a heavy step and it is commonly needed by multiple detectors, you are going to have repeat the post processing step on the "loader"  inside all  the detectors individually.

So to conserve space & time, my idea was to use the browser which calls "build_parallel" (where post processing can be done)  before injecting to all the detectors. Now looking back though, I feel like it was over engineering - maybe we don't need any post-processing of the ast.

The browser also provided a hook to the loader as an internal member. (which I didn't make clear :P  my bad ) 

So anyways I'll look more into this and get back to you Alex.

alexroan: @TilakMaddy There's no need to apologize! Those changes you made prompted me to dive deeper into this stuff. Check out these two PRs I've been working on today!:
1. [Extracting children](https://github.com/Cyfrin/aderyn/pull/120)
2. [Getting parents (sort of)](https://github.com/Cyfrin/aderyn/pull/121)


They're in `dev` and have not been released yet. Let me know your thoughts on them and any changes you think we should make.

PR #121, opened 2024-01-08
Browser: GetParent - enable detectors to easily get a parent of a node

### LINKED ISSUE #115: Enable detectors to "step out" of a node into a parent

alexroan:
Sort of Fixes #115 .

When the `ContextLoader` initially builds context, it keeps track of each Node's `SourceUnit`, `ContractDefinition`, `FunctionDefinition` and `ModifierDefinition`. The `parents` module then provides a framework to get a parent of these types.

For example, let's say we have an expression, and we want to get it's parent contract:

```rs
let contract_definition: &ContractDefinition = GetParent::contract_definition_of(my_expression, context_loader).unwrap();
```

Similarly, the extractor pattern can get the expressions from the contract definition (implemented in #120 ):

```rs
let expressions: Vec<Expression> = ExtractExpressions::from(contract_definition).extracted;
```

### DIFF
--- a/aderyn_core/src/context/browser/extractor.rs
+++ b/aderyn_core/src/context/browser/extractor.rs
@@ -11,7 +11,7 @@ pub struct ArrayTypeNameExtractor {
 }
 
 impl ArrayTypeNameExtractor {
-    pub fn from_node<T: Node + ?Sized>(node: &T) -> Self {
+    pub fn extract_from<T: Node + ?Sized>(node: &T) -> Self {
         let mut browser: ArrayTypeNameExtractor = Self::default();
         node.accept(&mut browser).unwrap_or_default();
         browser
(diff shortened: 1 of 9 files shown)


TilakMaddy:
This is stunning 🤩 !  Extractor pattern covering all types of nodes : ) When I create my own detectors I think I can envision its true potential.

TilakMaddy:
One thing that I can think of for now is that in `leptos` there was an issue where `cx` had to be passed down the component tree just to pass it back up and it was basically dead weight. That was removed in this PR https://github.com/leptos-rs/leptos/pull/918

So I could imagine something like that in the distant future where `loader` becomes  `cx` if we have more patterns like  ` GetParent::contract_definition_of(my_expression, context_loader)` introduced.

* Ack in release notes : https://github.com/leptos-rs/leptos/releases/tag/v0.5.0-beta

PR #122, opened 2024-01-09
make it easy for detectors to capture found instances of any AST type

TilakMaddy:
Before:
```rust
            self.found_instances.insert(
                loader.get_node_sort_key(&ASTNode::FunctionCall(item.clone())),
                item.src.clone(),
            );
```

After:
```rust
           capture!(self, loader, item); 
```

*  Done by adding `from`  methods on 'ASTNode` to allow any valid node to take the form of an ASTNode

*  Macro introduced that will automatically do everything for you so long as `found_instances` is declared in the first argument's fields (here it is `self`)


TilakMaddy:
Wait ... report markdown hasn't changed but json has ? strange

aderyn_core/src/context/mod.rs
```diff
@@ -1,2 +1,12 @@
 pub mod browser;
 pub mod loader;
+
+#[macro_export]
+macro_rules! capture {
+    ($self:ident, $loader:ident, $item:expr) => {
+        $self.found_instances.insert(
```
alexroan, line 7: Since each detector can define its struct, `found_instances` may be named something different in each one.

Is there a way that we can abstract/standardize this struct so that it's always the same and the detectors don't need to define it?

Also, the return value of `detect` and the `instances` function are related to this. Is there a way to abstract all of these away and have macros do this capturing/returning behind the scenes?

The user persona that I'm thinking about is someone who is writing detectors and nothing else. Can we make their job as easy as possible?

alexroan (commented):
Love the idea, but some refinement required.

aderyn_core/src/context/mod.rs
TilakMaddy: Hi Alex!

> Is there a way that we can abstract/standardize this struct so that it's always the same and the detectors don't need to define it?

Yes, I have an idea to standardize it but abstracting that away would be tricky (at least for me).  Also I'd like to not lean towards procedural macros simply because they are a pain to work with and just less portable in general. `capture!` is a declarative macro which is light in weight and can prompt the user to create a variable `found_instances` if not found.

> The user persona that I'm thinking about is someone who is writing detectors and nothing else. Can we make their job as easy as possible, without requiring them to adhere to implicit rules (like the naming of struct variables), and reduce the number of footguns?

Yes, I was thinking about this too! : )
So let's actually start thinking about distinguishing between user experience for bot developers who are going to create their own detection modules and aderyn contributors separately

When it comes to the former, I would like to make it such that they don't have to setup this heavy aderyn codebase just to write and submit detctors. Their codebase look like this

[https://github.com/TilakMaddy/custom_detector_example](https://github.com/TilakMaddy/custom_detector_example)

Please 🙏  set it up locally and go through it. (is linked to my other PR as well) See how lightweight it feels and look at the `cargo.toml` and `main.rs` 

If you feel that's a good start, then we can have `aderyn_pilot` to do the following

`aderyn_pilot new my_contest_bot` (Creates files like cargo new  + add aderyn dependencies and boilerplate)

`aderyn_pilot generate mev_detector ` (Creates a detector file with all the boilerplate as seen in my repo like struct name MevDetector  and member variable `found_instances` + `instances()` function + empty test suite)

`aderyn_pilot run <detector_name>` etc, etc (I'm still in the planning phase :P )

See this way, I believe users will feel more welcome
Maybe we could even have something  like `aderyn_pilot submit bot` that can sign the commit has with the private key and send it over to codehawks and stuff like that .... Later users can open source their repo which can be considered for merging the detector into the aderyn core

alexroan (approved):

PR #123, opened 2024-01-09
Feature/spin out detectors - Other codebases can subscribe to + write custom detectors

TilakMaddy:
### What this PR does

* External Rust codebases can subscribe to custom detectors  by installing `aderyn_driver` packages
  [My personal repo that demonstrates this](https://github.com/TilakMaddy/custom_detector_example)
 
  -> look inside  it only requires to install aderyn_driver as a high level dependency
  -> 3 variants showcased inside`main.rs` (default + subscription  + custom with subscription hyrbid)

### How it does it

* Decouples `run_with_printer`  which has `get_all_detectors` baked in. This is exploited so that we can have our own detectors passed in
* Re-exports some of the modules `aderyn_core` in `aderyn_driver` so that users need not install 2 different packages separately

Other

* Fixes Python Bindings for exclude and no_snippets flag
* Depends on `capture!` macro used in the previous PR


aderyn_core/src/context/loader.rs
```diff
@@ -58,6 +58,618 @@ pub enum ASTNode {
     WhileStatement(WhileStatement),
 }
 
+impl From<ArrayTypeName> for ASTNode {
```
alexroan, line 61: I merged the previous PR #122, and the diff here is similar. Could you rebase this branch based on the latest `dev` branch so these repeated changes aren't present?

TilakMaddy: Done Alex! 👊

alexroan:
This is a huge win, so I'd like to give it its dedicated release (maybe v0.0.12).

Could you include a test codebase here? I'd love to have this external detector functionality tested within this repo so that we know if/when we break it. Could you move your personal extension repo into this in a new directory and write tests for it?

alexroan:
This will need a rebase after the latest release! (sorry @TilakMaddy )

TilakMaddy:
Hey @alexroan ! I have made the necessary changes. Please track it on https://github.com/Cyfrin/aderyn/pull/128

TilakMaddy:
> This will need a rebase after the latest release! (sorry @TilakMaddy )

Heii that's an opportunity for me to keep up with the changes : ) In the beginning I learned a lot about aderyn when solving merge conflicts :P

PR #128, opened 2024-01-10
Feature/spin out detectors

TilakMaddy:


bot_reports/org_custom_subscription_analysis_report.md
```diff
@@ -0,0 +1,132 @@
+# Aderyn Analysis Report
```
alexroan, line 1: Should we add a cli script to make generating these easier?

bot_example/src/unindexed_events.rs
```diff
@@ -0,0 +1,84 @@
+use std::{collections::BTreeMap, error::Error};
+
+use aderyn_driver::context::loader::ContextLoader;
+use aderyn_driver::detection_modules::capture;
```
alexroan, line 4: Screenshot 2024-01-10 at 13 41 34

Everything builds fine, but I'm getting this in my vscode. The same with `capture!`. Do you get the same?

alexroan (commented):
This is looking really good.

bot_example/src/unindexed_events.rs
```diff
@@ -0,0 +1,84 @@
...
+#[cfg(test)]
+mod unindexed_event_tests {
+
+    use super::UnindexedEventsDetector;
+    use aderyn_driver::detector::detector_test_helpers::load_contract;
+    use aderyn_driver::detector::{Detector, IssueSeverity};
+
+    #[test]
+    fn test_unindexed_events() {
+        let context_loader = load_contract(
```
alexroan, line 64: Could we create another helper function that enables developers to run a test from source (either defined here in the rust file or in a file in this `bot_example` repo) rather than using the test files already present in the core repo?

alexroan (approved):

PR #133, opened 2024-01-24
`aderyn_pilot` -> `nyth`

alexroan:
* `aderyn_pilot` -> `nyth` (means nest in Welsh)
* `bot_starter_pack` -> `bot`. More intuitive navigation.
* add `foundry_workspace` to `bot`. Easier test setup for detector devs.


TilakMaddy:
Off the top of my head Alex what do you think of just `bot` instead of `bot_detectors` ? It's kind of understood there are detectors inside .

TilakMaddy:
Another thought - may be invalid :p I'm yet to go through the changes (I'll get back to you again)

Can we have a test in gh actions that makes sure that when you run `nyth new xxx` the Cargo toml inside the unpacked archive actually has aderyn driver linked via version and not rel.path ?

Basically some mechanism to make sure that when you publish nyth, it's archive doesn't outpit Cargo.toml with a relative path to aderyn_driver ?

You might have already taken care of it, I just wanted to make you aware ☺️

TilakMaddy:
> Another thought - may be invalid :p I'm yet to go through the changes (I'll get back to you again)
> 
> Can we have a test in gh actions that makes sure that when you run `nyth new xxx` the Cargo toml inside the unpacked archive actually has aderyn driver linked via version and not rel.path ?
> 
> Basically some mechanism to make sure that when you publish nyth, it's archive doesn't outpit Cargo.toml with a relative path to aderyn_driver ?
> 
> You might have already taken care of it, I just wanted to make you aware ☺️

Or

 we should have a separate gh actions for master branch that runs the prod version of bot fw (framework) assembler just before publishing . This way we can guarantee Cargo toml has the latest driver linked via version 

So i guess manually doing cargo publish may pose a threat...

TilakMaddy (approved):
This is looking really nice : ) I have tried it out. Only thing we need to be careful about like I said, is the archive.zip needs to be generated via 

`cargo run --bin bot_fw_assembler -- prod` 

which is unlike the `dev` version that we use in our gh actions. (We don't use prod because it should not build coz the version will always point to the latest unreleased version)

However when publishing we need to publish the prod version and make sure to do it after publishing aderyn_driver lib

alexroan:
Agreed! :) 

Regarding bot_archivegen, I agree, I've noticed some manual steps required whilst working on this PR.

Is there a way we can create a pre-publish step for nyth that runs the archivegen and checks versions? It's worth investigating!

PR #136, opened 2024-01-25
Fix: External detectors couldn't run a match case on ExtractedXXX::from(YY).extracted because AST node structs weren't available

TilakMaddy:
Solution: Re-export core ast into aderyn_driver for external detectors for ease of access.
An example detector that requires this ast node matching has been added to `bot_example`  to demonstrate the same.

Other solution:
Of course, the users can always cargo add `aderyn_core` for doing that but it makes it more convenient if we re-export this one more module in aderyn_driver via `core_ast`


bot_example/src/delegate_call_in_loop.rs
```diff
@@ -0,0 +1,95 @@
...
+    found_instances: BTreeMap<(String, usize), String>,
+}
+
+impl Detector for DelegateCallInLoopDetector {
+    fn detect(&mut self, loader: &ContextLoader) -> Result<bool, Box<dyn Error>> {
+        let mut member_accesses: Vec<MemberAccess> = vec![];
+
+        // Get all delegatecall member accesses inside for statements
+        member_accesses.extend(loader.for_statements.keys().flat_map(|statement| {
+            ExtractMemberAccesses::from(statement)
```
alexroan, line 22: Does the parent finder work here? `source_unit_of`, `contract_definition_of`, etc?

alexroan (commented):
Great catch, @TilakMaddy , one question.

bot_example/src/delegate_call_in_loop.rs
TilakMaddy: Yes ! `GetParent::XX` works - just tested it! 

All the stuff from the `browser` module is visible : )

alexroan (approved):
@TilakMaddy happy to merge once conflicts are resolved :)

alexroan (approved):

Issue #141, opened 2024-01-26
add scopefile option


PatrickAlphaC:
Instead of passing in a comma separated list into the CLI, it would be nice to read from a scopefile where the newline character acted as the comma. Let me know if this is not clear.

TilakMaddy:
https://github.com/Cyfrin/aderyn/pull/142

PR #142, opened 2024-01-27
Introducing aderyn.config.json

TilakMaddy:
If a file called `aderyn.config.json` is present in the root  project, aderyn users can specify what detectors should be run

```json
{
  "use_detectors": "push-zero-opcode,require-with-string,ercrecover"
}
```

Note: you can also pass the path to such a file to `aderyn` via `--config-file path/to/aderyn.config.json`

Now, if you want to know what values to pass in,

you must run

`aderyn registry`

which would look outptut something like this

```bash
Detector Registry

ID                               Title

delegate-call-in-loop          - Using `delegatecall` in loop
centralization-risk            - Centralization Risk for trusted owners
solmate-safe-transfer-lib      - Solmate's SafeTransferLib does not check for token contract's existence
avoid-abi-encode-packed        - `abi.encodePacked()` should not be used with dynamic types when passing the result to a hash function such as `keccak256()`
ercrecover                     - `ecrecover` is susceptible to signature malleability
deprecated-oz-functions        - Deprecated OpenZeppelin functions should not be used
....

```

And If you want to know more about a certain detector

you should run 

`aderyn registry <detector_id>` 

which would output something like this

```bash
Detector delegate-call-in-loop

Title
Using `delegatecall` in loop

Description
When calling `delegatecall` the same `msg.value` amount will be accredited multiple times.
```


TilakMaddy:
Oh and btw, you no longer have to `aderyn .`  You could just do `aderyn` in the working directory of the root

TilakMaddy:
BTW, you could also optionally add 

```json
{
   "scope_file" : "path/to/scopefile
}
```

Path will be relative to the location of `aderyn.config.json`

If scope is given in CLI argument as well as the config file, CLI will override config file

aderyn_core/src/detect/detector.rs
```diff
@@ -53,6 +57,54 @@ pub fn get_all_detectors() -> Vec<Box<dyn Detector>> {
...
+        "require-with-string".to_string(),
+        "non-reentrant-before-others".to_string(),
+        "block-timestamp-deadline".to_string(),
+        "unsafe-erc721-mint".to_string(),
+        "push-zero-opcode".to_string(),
+        "arbitrary-transfer-from".to_string(),
+    ]
+}
+
+pub fn get_detector_by_name(detector_name: &str) -> Box<dyn Detector> {
```
alexroan, line 83: We must alter these two functions each time we add/remove a dtector implementation. Is there a way in which we can achieve the same functionality without that requirement?

tests/aderyn.config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "use_detectors": "push-zero-opcode,useless-public-function",
```
alexroan, line 2: Could we make this an array instead of a string with comma-delimited input?

```json
"use_detectors": [
   "push-zero-opcode",
   "useless-public-function"
]
```

TilakMaddy:
Pausing to make this type safe

tests/aderyn.config.json
```diff
@@ -0,0 +1,4 @@
...
+    "scope_file": "../tests/scope.txt"
```
alexroan, line 3: I'm toying with the idea of defining the scope in this config file instead of a separate one. What do you think @TilakMaddy ?

alexroan (commented):
Cool feature! I've left some comments @TilakMaddy

alexroan:
This fixes #141

tests/aderyn.config.json
TilakMaddy: we can have both : )

scope_file ... (like it is right now)
scope .. (like you mentioned)

tests/aderyn.config.json
TilakMaddy: Ah okay! I like that too : ) but more to type tho .  People will be more used to this because slither :P What do you think ?

aderyn_core/src/detect/detector.rs
TilakMaddy: Now we have Enums that can derive kebab-case strings : ) 

(The only reason I'm not including hashmap is because I'd then have to force the bot writers to make their detector structs Cloneable because how dyn-traits work :P )

tests/aderyn.config.json
TilakMaddy: also if it's a scope file it can be anything - people would just have to drop in this config. So it's more like we adjust for them rather than they adjust for us : )

TilakMaddy: let's have a separate PR for that : ) This is more for like registry and introducing aderyn.config.json : )

TilakMaddy:
Update: detector names now get output in the json

aderyn_core/src/detect/detector.rs
```diff
@@ -42,6 +49,73 @@ pub fn get_all_detectors() -> Vec<Box<dyn Detector>> {
...
 
+pub fn get_all_detectors_names() -> Vec<String> {
+    get_all_detectors().iter().map(|d| d.name()).collect()
+}
+
+// Note to maintainers: DO NOT CHANGE THE ORDER OF THESE DERIVE ATTRIBUTES
+#[derive(Debug, PartialEq, EnumString, Display)]
+#[strum(serialize_all = "kebab-case")]
+pub(crate) enum DetectorNamePool {
+    DelegateCallInLoop,
```
alexroan, line 60: Hmmmm, on second thoughts @TilakMaddy , do you think this Enum adds too much back-and-forth between here and detectors?

TilakMaddy: Yes, Alex! But it's when a new detector gets formalized to aderyn codebase - we'd have to create a new entry in the detector pool to return in the `name()` function and the rust compiler will guide us to make the remaining changes.

Maybe not back & forth ...more like back once, forth from then on : ) :P

TilakMaddy:
Update:  @alexroan  On second thoughts I realized it is not much of a change to support scope as well as scope_file. So I have made the changes here : ) 

Change summary

```javascript
{
 scope: ["A.sol", "B.sol"],
scope_file: "scope.txt"
}
```
Both scope as well as lines inside scope_file will be concatenated : )

TilakMaddy:
I'll add more tests later and then undraft it : )

alexroan (approved):

PR #148, opened 2024-01-31
Naming and Typing changes

alexroan:
Comprised of two main changes:
* All references to an AST Node's ID now uses the `NodeID` type instead of `i64`
* `ContextLoader` has been renamed to `WorkspaceContext` for more accurate description of what it is.
  * Also, all references to "loader" have been replaced with "context"


alexroan:
cc: @TilakMaddy

Issue #146, opened 2024-01-29
Discussion: Expand the definition of a detector


alexroan:
Currently, the definition of a detector is a piece of code that finds and stores issues and makes them available to the report writer.

Following a discussion with [Chase](https://twitter.com/ChaseTheLight99), this idea is to broaden that definition so that detectors are reusable and can do more granular, repeatable things without printing the results to a report.

For example, a detector could be written solely to find all instances where a state variable was altered (feed it a state variable VariableDeclaration and find all Assignments with it on the leftHandSide). This detector would be called by other detectors rather than by the aderyn driver, so it would not produce items for the report.

We could implement this sort of reusable feature in the `browser` package, which currently hosts the Extractor and Parent builtin functionality. However, that locks detector developers into only the functionality in the core code. Now that we have the `nyth` development framework available, it would be great if we could leverage the existing concepts (detectors) and tooling to produce the same outcome, enabling developers to create their own "builtins" whilst they work on their reporting detectors.

There aren't a huge amount of changes to achieve this. The ones that come to mind are:
* Change `found_instances` to have a value of NodeId (i64), change macro
* Add an extra parameter to `Detector.detect` for reusable purposes, ignore for normal detectors?
* Unkown unknowns (I'll look into these ASAP)

The other options are:
* Introduce a new concept, separate from `detectors`, that covers this. Add the new concept to `nyth` with a new command and teach what it's for.
* Continue to build out the built-ins (Extractors) without any of this stuff. (<- This is not really an option if we want to leverage the community to rapidly write useful detectors, IMO).

cc: @TilakMaddy

TilakMaddy:
Off the top of my head (not describing a solution yet)
.
.
.

Bot writers would have more of an interest in creating the bot that works to produce results. Not necessarily maintaining coding best practices and ensuring scalability via reusable components.


This can imply the following

a) they may not even use even the extractxx or getparentxx in the first place. (They may just manually do it )

Surely we can recommend not denying that...

b) coming to your first set of points - yes i think it's good to encourage bot writers to creating more code structures like let's say a helper detector that has a detect function where instances() will only yield the node ID rather than the whole line number thing and stuff...and we can leverage nyth to do that ...when it comes to formalizing it in aderyn we can see where it best fits in the browser package and update docs

TilakMaddy:
Second thoughts 
------------------

In ReactJS, we have class components and function components (maybe class isn't very popular now but eh just for the sake of drawing a parallel)

[...]

PR #149, opened 2024-01-31
Option 1: Optional detect parameters

alexroan:
Introduce a new Detector type: `Utility`. The intention is for these utility detectors to be reusable, rather than produce report output.

With this, two new optional params are passed into `Detector::detect`:
* `within: &[NodeID]` - For the detector to search _within_, if passed (the default is the entire workspace_context)
* `using: &[NodeID]` - The specific search parameters

An example is created here to retrieve all instances where identifiers reference a state variable passed into `using`.

This is an implementation of the discussion: #146


alexroan:
@TilakMaddy Please review and let me know what you think.

aderyn_core/src/detect/detector.rs
```diff
@@ -119,6 +119,7 @@ pub fn get_detector_by_name(detector_name: &str) -> Box<dyn Detector> {
 
 #[derive(Debug, PartialEq)]
 pub enum IssueSeverity {
+    Utility,
```
alexroan, line 122: Should we call this a different name? (cc: @ChaseTheLight01, @TilakMaddy )

TilakMaddy: I haven't gotten a chance to review this wholly but I was wondering if it's possible to have a separate trait itself like DetectorUtility or something with its own severity enumerations (or maybe verbosity enumerations) ....

TilakMaddy: in the future, we would have separate detector-esque traits anyways for vyper ast . So.

TilakMaddy: It may make it harder for us to manage issue severities across all detector traits

TilakMaddy:
Another reason to have a separate detector trait - 

In our get all detector functions, we can continue to expose all detectors, not have to worry about is it's reporting detector or a utility detector.

TilakMaddy:
However I did get the gist of this though - what we are really looking for is a means to pass "messages" amongst detectors so that one can rely on the work of the other

-   -  -  -   - 
More gibberish from me 

 Right now it is 'within' and 'using'. If we choose to to add more params in the future without encompassing them in a struct like DetectorHelperArgs it may be hard to manage

aderyn_core/src/detect/util/identifiers_that_reference_state_variables.rs
```diff
@@ -0,0 +1,115 @@
...
+    found_instances: BTreeMap<(String, usize), NodeID>,
+}
+
+impl Detector for IdentifiersThatReferenceStateVariableDetector {
+    fn detect(
+        &mut self,
+        context: &WorkspaceContext,
+        _: &[NodeID],
+        using: &[NodeID],
+    ) -> std::prelude::v1::Result<bool, Box<dyn Error>> {
```
TilakMaddy, line 23: Please use just `Result`

TilakMaddy: If we have a `bot_example` that uses this Detector it would be nice : )

TilakMaddy (changes requested):

aderyn_core/src/detect/detector.rs
alexroan: Yes, that makes sense!

Ooooo, I like the idea of a separate impl. Within `nyth` we can then have different commands for creating report detectors vs reusable detectors

aderyn_core/src/detect/util/identifiers_that_reference_state_variables.rs
alexroan: Agreed.

alexroan:
Closing this in favour of #150

aderyn_core/src/detect/detector.rs
TilakMaddy: Creating a generate subcommand for nyth to add resuable detectors 

Todo : Add to Milestone

PR #150, opened 2024-02-02
Option 2: Reusable detector trait

alexroan:
Introduce a new Detector trait: `ReusableDetector`. The intention is for these detectors is to be reusable, rather than produce report output.

The only function within `ReusableDetector` is `detect`

`ReusableDetector::detect` differs from `Detector::detect`:
* `workspace_context: &WorkspaceContext` - 
* `within: &[NodeID]` - For the detector to search _within_, if passed (the default is the entire workspace_context)
* `using: &[NodeID]` - The specific search parameters

and returns an array of NodeIDs.

An example is created here to retrieve all instances where identifiers reference a state variable passed into `using`.

This is an implementation of the discussion: #146


aderyn_core/src/detect/detector.rs
```diff
@@ -168,6 +168,17 @@ pub trait Detector {
     }
 }
 
+pub trait ReusableDetector {
+    fn detect(
```
alexroan, line 172: @TilakMaddy separate trait with a single function: `detect`

aderyn_core/src/detect/reusable/identifiers_that_reference_a_function.rs
```diff
@@ -0,0 +1,67 @@
+use crate::{context::workspace_context::ASTNode, detect::detector::ReusableDetector};
+
+#[derive(Default)]
+pub struct IdentifiersThatReferenceAFunctionDetector {
+    // All the state variables, set at the beginning of the detect function
+    found_instances: Vec<ASTNode>,
+}
+
+impl ReusableDetector for IdentifiersThatReferenceAFunctionDetector {
```
alexroan, line 9: Example reusable detector

aderyn_core/src/detect/nc/useless_public_function.rs
```diff
@@ -19,23 +19,17 @@ pub struct UselessPublicFunctionDetector {
 
 impl Detector for UselessPublicFunctionDetector {
     fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
```
alexroan, line 21: A normal detector that uses a reusable detector @TilakMaddy

aderyn_core/src/detect/reusable/identifiers_that_reference_a_function.rs
```diff
@@ -0,0 +1,67 @@
...
+        if using.len() > 1 {
+            return Err("Error: using is greater than 1".into());
+        }
+
+        let retrieved = &using[0];
+        if let ASTNode::FunctionDefinition(function_definition) = retrieved {
+            context.identifiers.keys().for_each(|identifier| {
+                if identifier.referenced_declaration == function_definition.id {
+                    self.found_instances
+                        .push(ASTNode::Identifier(identifier.clone()));
```
TilakMaddy, line 28: Can be changed to `self.found_instances.push(identifier.into());`

aderyn_core/src/detect/nc/useless_public_function.rs
```diff
@@ -19,23 +19,17 @@ pub struct UselessPublicFunctionDetector {
...
-                f.visibility == Visibility::Public
-                    && f.kind != FunctionKind::Constructor
-                    && !referenced_functions.contains(&f.id)
-            })
-            .collect::<Vec<_>>();
+        let unreferenced_public_functions =
+            context.function_definitions.keys().filter(|&function| {
+                matches!(function.visibility, Visibility::Public)
+                    && IdentifiersThatReferenceAFunctionDetector::default()
+                        .detect(
```
TilakMaddy, line 26: Can be changed to 
`.detect(context, &[function.into()], &[])`

aderyn_core/src/detect/reusable/identifiers_that_reference_a_function.rs
```diff
@@ -0,0 +1,67 @@
...
+        let function_definition = context
+            .function_definitions
+            .keys()
+            .find(|function_definition| function_definition.name == "increment")
+            .unwrap()
+            .clone();
+
+        let mut detector = IdentifiersThatReferenceAFunctionDetector::default();
+        // create vec with function_definition as item 0
+        let using = vec![ASTNode::FunctionDefinition(function_definition)];
```
TilakMaddy, line 62: same pattern of `into` maybe used here as well

TilakMaddy (changes requested):
This is looking good 😊 : ) Just some semantic changes requested (because my guess is people would look through the existing code bases to learn to write detectors)

Other than that, my last comment would be if we're unsure how well things are we can try to re-export those `resuable` modules under `detection_modules::alpha::v1` namespace so then people will know this is experimental


Thank you : )

PR #159, opened 2024-02-14
reduce steps in integration tests

TilakMaddy:
I had these written out before having unit tests for lightchaser. Now, I don't think having so many of them serves any purpose. We can get rid of them  to make `cli/reportgen.sh` finish running sooner

Other improvement

* Added `cli/judgeops.sh` so that it acts like an alias. (We would have to interact with judgeops from the root directory anyways.  So using this can help reinforce that)

Example: `./cli/judgeops.sh --help` instead of `cargo run --bin judgeops -- --help`


TilakMaddy:
@alexroan resolved merge conflicts here. So pls get this in next

TilakMaddy:
Outdated

PR #166, opened 2024-02-18
Extract Parent chain of an AST Node (all the way upto contract definition)

TilakMaddy:
As pointed out in the below comment, you can make use of `GetParentChain::of(xxx)` to traverse the AST upward.


aderyn_core/src/detect/experimental/demo_parent_chain.rs
```diff
@@ -0,0 +1,126 @@
...
+#[derive(Default)]
+pub struct ParentChainDemonstrator {
+    // Keys are source file name and line number
+    found_instances: BTreeMap<(String, usize), NodeID>,
+}
+
+/*
+
+In ParentChainContract.sol, there is only 1 assignment done. The goal is to capture it first.
+Then walk up it's parent tree and capture a) contract definition b) if statement c) for statement
```
TilakMaddy, line 23: This file shows an example of how you could traverse the AST upward from a given node

tests/contract-playground/src/ParentChainContract.sol
```diff
@@ -0,0 +1,48 @@
...
+    uint256 public number = 0;
+
+    function setNumber(uint256 newNumber) public {
+        
+    }
+
+    function increment(uint256 newNumber) public {
+        if (number < 5 && number > 1) {
+            for (uint256 j = 9; j != 0; --j) {
+                number = newNumber;
```
TilakMaddy, line 18: We try to find this assignment first. Then walk up the tree to its contract definition

aderyn_core/src/context/browser/chain.rs
```diff
@@ -0,0 +1,580 @@
...
+        get_parent_chain_of_child(self.id, context)
+    }
+}
+impl GetParentChain for WhileStatement {
+    fn of(&self, context: &WorkspaceContext) -> Vec<ASTNode> {
+        get_parent_chain_of_child(self.id, context)
+    }
+}
+
+pub fn get_parent_chain_of_child(child_id: NodeID, context: &WorkspaceContext) -> Vec<ASTNode> {
```
TilakMaddy, line 223: Heart of the logic for this PR

TilakMaddy: This function also serves the next PR https://github.com/Cyfrin/aderyn/pull/167 which allows us to get immediate children

TilakMaddy:
I'm not too happy with how I named stuff tbh. In your review for this PR, I'd love to see naming suggestions included as well.

TilakMaddy:
@alexroan 

#### Small Explanation of how this is accomplished

* `F1` : Introduced `ExtractEverything` API which accumulates the output from all of the existing `ExtractXXXX` API
* `F2` : Introduced a function that will accumulate all the `ASTNode` from the `WorkspaceContext`
* Let's say `T` is the node for which you want to find the parent/ancestry chain.
* Then we do the following
 ```python

ancestors = []

for node in F2():
    for child in F1(node): # F1 does not return not just immediate children, all children
         if child.id == T.id: # If yes, child is an ancestor of T
              ancestors.push(child)

# Now, we have collected all ancestors of T, so it's a matter of ordering them.

chain = []

for ancestor in ancestors:
    child_count = len(F1(ancestor))
    chain.push((child_count, ancestor))

chain.sort() # This will arrange ancestors in increasing order of their number of children

# We also know that a grandparent will always have more total children (not just immediate) than just the parent

# Hence, we can conclude `chain` contains parents in the order of closest to farthest

chain = map(lambda x: x[1], chain)        

RETURN chain

```

\* Please let me know if more is required

File: aderyn_core/src/context/browser/chain.rs (hunk 3 above)
Comment (alexroan) [reply aderyn_core/src/context/browser/chain.rs:223]: My concern with this being here is that this work is done every time a detector calls `GetParent::of`. As the set of detectors grows, the number of calls to this will increase. Obviously, we can introduce caching and other tricks to help the time complexity, but it still feels like an antipattern to do all of this work _each time_.

I wonder if a preprocessing step can prepare the ASTNodes to have a `.parent()` accessor. The preprocessing runs only once when the contract AST files are loaded initially (the entire tree is traversed originally _anyway_). That feels like the more efficient and clean way to do this. Can we piggyback off the initial tree traversal of Aderyn? @TilakMaddy

File: aderyn_core/src/detect/experimental/demo_parent_chain.rs [hunk 4]
```diff
@@ -0,0 +1,126 @@
...
+
+*/
+
+impl IssueDetector for ParentChainDemonstrator {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
+        for assignment in context.assignments.keys() {
+            // Retrieve Parent Chain (from closest to farthest)
+            let parents = GetParentChain::of(assignment, context);
+            {
+                println!("Parent Chain (from closest to farthest)\n---------");
```
alexroan, line 34: I'm trying to think of what I would prefer:
1. This functionality (chain in a vector)
2. `.parent(),` with the ability to do. `parent().parent().parent()` if I want to go three levels up

To be honest, I think I prefer version 2, but I'm not sure yet @TilakMaddy

aderyn_core/src/context/browser/chain.rs
TilakMaddy: Thanks for the input : ) It definitely feels like an antipattern to do each time and. 

I think what we can do, is like you said preprocess the tree by creating a new hashmap field in workspace context that can give all the IDs of all children nodes for a given node ID.

I don't know if we can piggyback off the first traversals but definitely we can ensure that GetParent on a given node doesn't run twice

Let me try it out. In the meantime if you have any other ideas let me know.

aderyn_core/src/detect/experimental/demo_parent_chain.rs
TilakMaddy: @alexroan There's a PR for that already....

https://github.com/Cyfrin/aderyn/pull/168 !!

PR #173, opened 2024-02-22
Issue clarification PR

TilakMaddy:
@alexroan I'm a little confused here. Please let me know if I am missing something

Step 1
clone the dev branch

Step 2
delete the out folder in tests/contract-playground

Step 3
forge build 

Step 4
cli/reportgen

Step 5
cargo test. ---------> this fails !

PR #174, opened 2024-02-23
Update solc version of files after running foundryup

TilakMaddy:
As explained here, https://github.com/Cyfrin/aderyn/pull/173 the tests break

This happens due to 0.8.23.json files being formed rather than 0.8.21.json 

(remove `out` and `cache` folder inside `tests/contract-playground` and run `forge build`)

You will see that cargo test fails

This PR fixes it by renaming all references to *0.8.21.json files in the load contract tests to *.0.8.23.json

Other fixes
-----------
Cargo compile waning in `aderyn_core/src/detect/reusable/identifiers_that_reference_a_function.rs`


alexroan (approved):

alexroan:
This was reverted because it meant locally I would need to switch back manually.

I'm thinking it's an issue with differences in our foundry setup, which may be OS related?

PR #190, opened 2024-02-27
id() method  for AST Nodes.

TilakMaddy:
~~So if you ever wanted to "just" capture an ASTNode regardless of what internal type it is, then this PR will help you achieve that~~.

~~It can be used in the future when we have a GetParentChain and you want to capture up to 3 parents for context to see in the final report (i.e, father, grandfather and great grandfather nodes) then you may do so with `capture_raw!`~~

~~Now here we have a much simpler, silly example to demo the concept but the idea remains~~

`id() method on ASTNode`


aderyn_core/src/detect/mod.rs
```diff
@@ -1,4 +1,5 @@
 pub mod detector;
+pub(crate) mod experimental;
```
TilakMaddy, line 2: This should be pub(crate) because the driver (and therefore nyth) shouldn't see this module

aderyn_core/src/detect/mod.rs
```diff
@@ -14,4 +15,16 @@ macro_rules! capture {
     };
 }
 
+#[macro_export]
+macro_rules! capture_raw {
```
alexroan, line 19: Is there a way we can merge this logic into `capture` so we just have the one macro @TilakMaddy ?

alexroan:
This needs rebasing to the latest `dev` @TilakMaddy

report.md
```diff
@@ -686,6 +686,12 @@ Assigning values to address state variables without checking for `address(0)`.
 	    function foo2(address bar) public pure {
 	```
 
+- Found in src/cloc/HeavilyCommentedContract.sol [Line: 31](tests/contract-playground/src/cloc/HeavilyCommentedContract.sol#L31)
```
TilakMaddy, line 689: @alexroan  I just added an `id()` method on ASTNode... How come this is newly added to the report? (it's not like I changed a solidity file...) 👀

TilakMaddy:
I plan on using this `id()` method on ASTNode for easily forming the parent chain

report.md
TilakMaddy: UPDATE: (i found out)

Please run `forge build --ast` in tests/cotract-playground before running `cli/reportgen.sh`.. You will see that these 2 were missing

alexroan (approved):

PR #193, opened 2024-02-28
Print src in json output

alexroan:
More detail for json output 👀


report.md
```diff
@@ -829,12 +895,30 @@ Assigning values to address state variables without checking for `address(0)`.
...
+	        ret[1] = bytesValue[6];
+	```
+
 - Found in src/eth2/DepositContract.sol [Line: 157](tests/contract-playground/src/eth2/DepositContract.sol#L157)
 
 	```solidity
 	        ret[2] = bytesValue[5];
 	```
 
+- Found in src/eth2/DepositContract.sol [Line: 157](tests/contract-playground/src/eth2/DepositContract.sol#L157)
```
alexroan, line 910: This seems to be adding duplicates where the issue is found multiple times on the same line.

alexroan: This is probably _good_, but there should be a way of displaying which characters on that line have caused the actual issue.

Also, this particular detector is crap. It should probably not have such a high hit rate, maybe only use it when the magic number is:

* on the right hand side of an assignment (assigning a variable to the magic number)
* In some complex math (this one is probably more difficult to catch)

alexroan: ^ Follow up for me

PR #195, opened 2024-02-28
v0.0.16 Staging

alexroan:


alexroan:
@TilakMaddy I'm going to cut the release here so I can test some of this stuff out with the VS code extension.

I'll review all the new detectors you've been creating after that 😍

PR #198, opened 2024-02-29
Introducing Comment Driven Testing for detectors

TilakMaddy:
This is to help with testing detectors (completely optional to use - no NEED of any changes in the existing workflow)

Problem:
we very comfortably use `assert detector.instances.len() == XX`  everywhere in tests because it's easy. We compromise on checking that the instances are indeed what we think they are... because too much code to write..

Yes I agree

Therefore, instead of code, you should mark in solidity file itself, what line you want to capture in the form of a comment

```solidity
// @nyth:blame(detector-name) <--- write this one line above what you want to capture
if (number > 3) {  
```

then in the tests,

do this

```rust
assert(detector.verify_blame_coverage(
 "path/to/solidityfile"
)) 
```

This will check if all the commented lines or I should say "blamed" lines are captured inside `found_instances` of that particular detector

See the below example + test for more details


tests/contract-playground/src/parent_chain/ParentChainContract.sol
```diff
@@ -13,7 +13,9 @@ contract ParentChainContract is Ownable, ReentrancyGuard {
     }
 
     function increment(uint256 newNumber) public {
+        // @nyth:blame(constants-instead-of-literals)
```
TilakMaddy, line 16: example here

tests/contract-playground/src/parent_chain/ParentChainContract.sol
```diff
@@ -13,7 +13,9 @@ contract ParentChainContract is Ownable, ReentrancyGuard {
...
         if (number < 5 && number > 1) {
+            // @nyth:blame(constants-instead-of-literals)
```
TilakMaddy, line 18: another example

aderyn_core/src/detect/nc/constants_instead_of_literals.rs
```diff
@@ -88,4 +88,27 @@ mod constants_instead_of_literals_tests {
...
+        let found = detector.detect(&context).unwrap();
+        assert!(found);
+
+        println!("{:?}", detector.instances());
+
+        // assert that the detector finds the correct number of instances
+        assert_eq!(detector.instances().len(), 3);
+
+        assert!(detector
+            .verify_blame_coverage(
```
TilakMaddy, line 109: here's where i check for those

tests/contract-playground/src/parent_chain/ParentChainContract.sol
alexroan: What happens when there are multiple instances on the same line (like this one 👀 )?

TilakMaddy: great point ! Should we  make it something like

`@nyth:blame(constants-instead-of-literals:2)`

???

TilakMaddy: ( to verify that it was blamed twice )

TilakMaddy:
Update:

NOTE: Look for pattern `@nyth:blame(detector-1-name:Y,detector-2-name:Z)`
In the Solidity source code. When you come across one interpret it like follows:
```
Check detector-1-name captures next line Y no of times AND
Check detector-2-name captures next line Z no of times
```

Note: the detector can however capture more instances than what is in the comments. That's why the name is `verify_blame_coverage` . Only make sure that AT LEAST what's commented has made it into the `found_instances`

tests/contract-playground/src/parent_chain/ParentChainContract.sol
```diff
@@ -13,7 +13,9 @@ contract ParentChainContract is Ownable, ReentrancyGuard {
     }
 
     function increment(uint256 newNumber) public {
+        // @nyth:blame(constants-instead-of-literals:2)
```
TilakMaddy, line 16: verify_blame_coverage makes sure that there are exactly 2 instances of next line captured

TilakMaddy: Now, there may be other lines that are captured, but since its not commented with `@nyth:blame` it doesn't matter . test still passes. 

so it's more like use this in places that you think are critical to make it in....

TilakMaddy: and you don't have to use this more than you'd like to use it in a codebase.... (basically `@nyth:blame(..)` need not be marked before every single instance that's captured)

TilakMaddy:
I'll fix this later ...

alexroan:
I like this idea, but closing for now due to being out of date with the `dev` branch.

A solution I found to this in the meantime can be seen in the tests for the [zero_address_check.rs](https://github.com/Cyfrin/aderyn/blob/dev/aderyn_core/src/detect/nc/zero_address_check.rs#L209)

TilakMaddy:
Okay that's also good  : ) 

Downside that I can see (for future)

- when the structure of the contract changes/evolves you will have to update the logic in the tests as well
- Whereas  here- the comment will always be 1 line above whatever you want to capture

PR #203, opened 2024-03-01
Node Js Bindings

TilakMaddy:
**On the command line in the `aderyn_nodejs` folder, do the following**

`npm install`
`npm run build-release`

Now type `node` and go into the shell

```nodejs
> const aderyn_driver = require('.')
> aderyn_driver.drive("../../6-thunder-loan-audit", "thunder-report.md", false, [], [])
Detecting framework...
Framework detected: Foundry mode engaged.
Foundry root path: "../../6-thunder-loan-audit"
Foundry src path: "../../6-thunder-loan-audit/src"
Get Detectors
Running 18 detectors
Detectors run, processing found issues
Found issues processed. Printing report
Report printed to thunder-report.md
true
```

If you like it then run `npm publish` 

----------

Functions exported in JS:

* .drive()
* .drive_with()
* .get_all_issue_detectors_names()

Check that drive, drive_with function returns true in javascript.


aderyn_nodejs/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+        output: output.value(&mut cx),
+        exclude: parse(exclude, &mut cx),
+        scope: parse(scope, &mut cx),
+        no_snippets: no_snippets.value(&mut cx),
+    });
+
+    Ok(cx.boolean(true))
+}
+
+fn drive_with(mut cx: FunctionContext) -> JsResult<JsBoolean> {
```
TilakMaddy, line 25: order of arguments for `drive_with(...)`

aderyn_nodejs/src/lib.rs
```diff
@@ -0,0 +1,78 @@
+use aderyn_driver::{
+    detector::get_issue_detector_by_name,
+    driver::{self, Args},
+};
+use neon::prelude::*;
+
+fn drive(mut cx: FunctionContext) -> JsResult<JsBoolean> {
+    let root = cx.argument::<JsString>(0)?;
```
TilakMaddy, line 8: order of arguments for `drive`

TilakMaddy:
@alexroan  On a broader note, I think that the true power bindings helps when there is "to-and-fro" communication not just fire and forget. For example here we just want to generate json report and forget (until provoked again)

Reason why I said that is I feel like we can use all the helper detectors to make damn good inspector! (not just vulnerability reporter alebit that's the fuel)

Imagine hovering over a contract definition that has a long chain of parent contracts and it shows the list of storage variables in the order from most base to most child. 

So like design wise
 Reusables ===> For inspectors + For issue detectors
 IssueDetectors ====> For Vulnerability report

aderyn_nodejs/src/lib.rs
alexroan: Could we add some TS tests to this that run in CI to make sure reports are produced correctly @TilakMaddy ?

TilakMaddy:
I couldn't add a test in TS as I'm unable to get typescript to import from "." here, index.node. Would need some help : )

However, I have attached a d.ts file - if that works, then it's also great!

aderyn_nodejs/create_report.js
```diff
@@ -0,0 +1,12 @@
...
+aderyn_driver.drive("../tests/contract-playground", "report_jsbindings.json", false, [], [])
+
+aderyn_driver.drive_with(
+    "../tests/contract-playground", 
+    "report_jsbindings_config.json",
+    false,
+    [],
+    ["Counter.sol", "Contract.sol"],
+    ["push-zero-opcode","useless-public-function"],
+);
```
TilakMaddy, line 12: Basically CI will check that this file re-creates the same assets as normal aderyn cli would

aderyn_nodejs/@types/aderyn_nodejs.ts
```diff
@@ -0,0 +1,22 @@
+declare module 'aderyn_nodejs' {
+
+    export function drive(
+        root: string,
+        output: string,
+        no_snippets: boolean,
+        exclude: string[],
```
TilakMaddy, line 7: This file should help in TS linting...but don't know how to make it work without actually publishing the package

alexroan:
This is awesome dude! Published here for now: https://www.npmjs.com/package/aderyn_nodejs

alexroan (approved):

TilakMaddy:
> This is awesome dude! Published here for now: https://www.npmjs.com/package/aderyn_nodejs

Follow up https://github.com/neon-bindings/neon/issues/117

PR #207, opened 2024-03-05
node_type() method for ASTNode

TilakMaddy:
* To help with identifying ASTNode type 

(NodeType can be used as input  in finding closest parent of type X)

Other fixes

Spelling mistake fixed in `NodeType` enum


TilakMaddy:
@alexroan  `NodeType` seems to have more types than `ASTNode` has... Do you think we should just have another enum like `ASTNodeType` which will cover exactly what is available in `ASTNode` or should re just go with `NodeType` ?

alexroan (approved):

PR #213, opened 2024-03-10
New detector: Internal functions called only once can be inlined

TilakMaddy:
Internal functions called only once can be inlined :P


TilakMaddy:
ChatGPT wrote the contract btw : )

image

aderyn_core/src/detect/nc/useless_internal_function.rs
```diff
@@ -0,0 +1,92 @@
...
+
+impl IssueDetector for UselessInternalFunctionDetector {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
+        let internal_functions = context
+            .function_definitions()
+            .into_iter()
+            .filter(|&function| matches!(function.visibility, Visibility::Internal));
+
+        for internal_function in internal_functions {
+            if IdentifiersThatReferenceAFunctionDetector::default()
```
alexroan, line 28: You used the reusable detector 🚀 

I'm going to approve this as-is, but I remember you were trying out a `Helper` pattern instead of this reusable detector pattern. Do you think that would make this cleaner?

alexroan (approved):
LGTM, There's one follow up question in one of the review comments @TilakMaddy , but I'm going to merge this  :)

aderyn_core/src/detect/nc/useless_internal_function.rs
TilakMaddy: Yes ! Let me try it : )

PR #214, opened 2024-03-10
New Detector: public constants can be replaced with a single getter function as a tuple

TilakMaddy:
If there are multiple public constants in a contract, they could be replaced by a single public getter function that returns all the public constants as a tuple. This way you can mark the constants as private and save deployment cost


alexroan:
Do you have a reference for this one @TilakMaddy ?

TilakMaddy:
I don't remember where I read but basically the idea is when you make something is public , bytecode is generated to expose that which increases deployment cost. So if you get away with doing it only once for all constants why not...

When I find it, I'll definitely comment the link to this PR

TilakMaddy:
@alexroan I have slightly changed what we capture her

Before - we captured all constants that were public in a contract that had more than 1 public constant.

Now - we simply capture all the contracts that have multiple public constants

alexroan:
I'm still not sure about this as a detector. It's a deployment cost benefit, but I'm not sure it fits with the security-minded advice that Aderyn intends to give.

Gas golfing is undoubtedly an important step during development, but there's a current trend of seeing it as equally important to security, which it is not (hence why GAS severity is not included in Aderyn). 

I'm going to close this PR for now, but if we find justification in the future, we'll reintroduce it @TilakMaddy .

TilakMaddy:
Understood I'll keep that in mind : )

PR #216, opened 2024-03-10
New Detector: Empty block detector

TilakMaddy:
Capture blocks with no statements! 

Exceptions: constructors, receive, fallback function

So that includes normal functions, if statements, for loops, stuff like that


TilakMaddy:
This is so weird... I don't get any errors when I try to generate `report.md` locally but the CI fails ! Apparently there is an Issue instance with line_no = 0 looks like ... But unable to replicate this locally

TilakMaddy:
Hmm ... Could it do with us not forcing line-ending standards in this repo ?

PR #217, opened 2024-03-10
New Detector: Empty blocks of code should be removed

TilakMaddy:
Capture blocks with no statements!

Exceptions: constructors, receive, fallback function

So that includes normal functions, if statements, for loops, stuff like that.


tests/contract-playground/src/EmptyBlocks.sol
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: MIT
```
alexroan, line 1: I've added this test contract in here

report.md
```diff
@@ -867,7 +869,85 @@ This is a best-practice to protect against reentrancy in other modifiers.
...
+
+Consider removing empty blocks.
+
+- Found in src/AdminContract.sol [Line: 14](tests/contract-playground/src/AdminContract.sol#L14)
+
+	```solidity
+	    function someOtherImportantThing() external nonReentrant onlyOwner {
+	```
+
+- Found in src/EmptyBlocks.sol [Line: 6](tests/contract-playground/src/EmptyBlocks.sol#L6)
```
alexroan, line 882: Constructor, receive and fallback are all caught unintentionally here @TilakMaddy . Can you please fix?

alexroan (changes requested):
This one doesn't quite work as intended @TilakMaddy

report.md
TilakMaddy: Thanks so much for reviewing and pointing out ! I have an idea : ) Let me fix it !

aderyn_core/src/detect/nc/empty_blocks.rs
```diff
@@ -0,0 +1,88 @@
...
+    found_instances: BTreeMap<(String, usize, String), NodeID>,
+}
+
+impl IssueDetector for EmptyBlockDetector {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
+        for empty_block in context.blocks().iter().filter(|b| b.statements.is_empty()) {
+            if let Some(ASTNode::FunctionDefinition(f)) =
+                &empty_block.closest_parent_of_type(context, NodeType::FunctionDefinition)
+            {
+                // It's okay to have empty block if it's a constructor, receive, or fallback
```
TilakMaddy, line 26: @alexroan I fixed it by changing the above from `.parent(context)` to `.closest_parent_of_type(context, NodeType::FunctionDefinition)`

alexroan: We would probably want this to capture constructors with nested empty blocks. Does this do that?

TilakMaddy: At the moment as we have it, we don't capture any functions here instead we capture the starting `{` of the empty blocks so long as they are not inside a special function like constructor, receiver, fallback

TilakMaddy: Do you think we should change this to capture the functions ?

TilakMaddy: I have changed it to capture the functions now : ). I have also, added some more of my own test cases to the `EmptyBlocks.sol`

Follows this convention : 

```rust
// Good -> don't capture
// Bad -> capture
```

tests/contract-playground/src/EmptyBlocks.sol
```diff
@@ -0,0 +1,65 @@
...
+
+contract EmptyBlocks {
+    // Bad
+    constructor() {
+        emit EmptyEvent();
+        // Empty Block Below
+        {
+
+        }
+    }
```
TilakMaddy, line 28: Added extra

TilakMaddy:
@alexroan due to the MEP problem (Mutually exclusive pragma, yes i made it up :P) I had to erase the entire out folder, change the pragma versions of contracts that are MEP pairs and then regenerate everything.

This was important coz otherwise non empty blocks were also being captured.

report.md
```diff
@@ -867,7 +887,73 @@ This is a best-practice to protect against reentrancy in other modifiers.
...
+- Found in src/AdminContract.sol [Line: 14](tests/contract-playground/src/AdminContract.sol#L14)
+
+	```solidity
+	    function someOtherImportantThing() external nonReentrant onlyOwner {
+	```
+
+- Found in src/EmptyBlocks.sol [Line: 22](tests/contract-playground/src/EmptyBlocks.sol#L22)
+
+	```solidity
+	    constructor() {
```
alexroan, line 903: I think this should capture the nested empty block on line 24, instead of the constructor definition. What do you think?

```sol
    constructor() {
        emit EmptyEvent();
        // Empty Block Below
        {

        }
    }
```

alexroan (changes requested):
One small fix remaining (what to capture). Looking good :)

tests/contract-playground/src/cloc/HeavilyCommentedContract.sol
```diff
@@ -3,7 +3,7 @@
 *//*
     also normal
 */
-pragma solidity >=0.4.22 <0.6.0; 
+pragma solidity >=0.8.0 <0.8.25;
```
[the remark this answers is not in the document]
TilakMaddy: No particular reason to choose >=0.8.0 <0.8.25;  . ... I was just trying to get rid of mutually exclusive paired pragmas in the contract playground : )

tests/contract-playground/src/eth2/DepositContract.sol
```diff
@@ -9,7 +9,7 @@
 
 // SPDX-License-Identifier: CC0-1.0
 
-pragma solidity 0.6.11;
+pragma solidity 0.8.20;
```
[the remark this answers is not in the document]
TilakMaddy: Ah okay I didn't know that .... but again MEP no?

Here are some workaround options I can think of - 

* Move this contract to a new playground
* Remove this from scope in the current playground (of course allow 0.8.* contracts that are using it to continue to use it)

report.md
TilakMaddy: Easy. I'll do it right away !

TilakMaddy: For now I'm capturing Line 25 which where the `{` starts.

TilakMaddy:
Summary:

> If the empty block in present in a normal function, capture the function defn
> Else capture the start of the empty block (in case of constructor, fallback, receiver)

tests/contract-playground/src/EmptyBlocks.sol
```diff
@@ -0,0 +1,89 @@
...
+
+    // Bad
+    fallback() external payable {
+        // Empty block
+        {
+
+        }
+    }
+}
+
```
TilakMaddy, line 43: newly added

TilakMaddy:
> @alexroan due to the MEP problem (Mutually exclusive pragma, yes i made it up :P) I had to erase the entire out folder, change the pragma versions of contracts that are MEP pairs and then regenerate everything.
> 
> This was important coz otherwise non empty blocks were also being captured.

Please acknowledge this before merge : ) 

I had to do it in order to stop false nodes from being captured !

TilakMaddy:
Note -  I have merged cross compilation ci here only to get over the lints correctly :P  Hope it's okay

alexroan (approved):
:)

PR #219, opened 2024-03-14
Node IDs are  not unique - hence we lose captured occurrences due to replacement taking place in the `nodes` hashmap

TilakMaddy:
## Proof of bug

To see that in action, go into Workspace Context, and find `visit_literal`, then make the following changes.

```diff
fn visit_literal(&mut self, node: &Literal) -> Result<bool> {

-       self.nodes.insert(node.id, ASTNode::Literal(node.clone()));
+       let x = self.nodes.insert(node.id, ASTNode::Literal(node.clone()));
+       if x.is_some() {
+           let existing = x.unwrap();
+            if existing.src() != Some(&node.src) {
+                println!("Existing: {:?}\n", &existing);
+                println!("Replacing with: {:?}\n\n", node);
+           }
+      }

        self.literals_context.insert(
            node.clone(),
            NodeContext {
                source_unit_id: self.la
```

You will find that output looks like this 

```bash
Existing: Literal(Literal { hex_value: Some("333435"), value: Some("345"), subdenomination: None, kind: Number, argument_types: None, is_constant: false, is_l_value: false, is_pure: true, l_value_requested: false, type_descriptions: TypeDescriptions { type_identifier: Some("t_rational_345_by_1"), type_string: Some("int_const 345") }, src: "128:3:0", id: 3 })

Replacing with: Literal { hex_value: Some("30"), value: Some("0"), subdenomination: None, kind: Number, argument_types: None, is_constant: false, is_l_value: false, is_pure: true, l_value_requested: false, type_descriptions: TypeDescriptions { type_identifier: Some("t_rational_0_by_1"), type_string: Some("int_const 0") }, src: "166:1:0", id: 3 }


Existing: Literal(Literal { hex_value: Some("307835423338446136613730316335363835343564436663423033466342383735663536626564644334"), value: Some("0x5B38Da6a701c568545dCfcB03FcB875f56beddC4"), subdenomination: None, kind: Number, argument_types: None, is_constant: false, is_l_value: false, is_pure: true, l_value_requested: false, type_descriptions: TypeDescriptions { type_identifier: Some("t_address"), type_string: Some("address") }, src: "180:42:0", id: 6 })

Replacing with: Literal { hex_value: Some("30"), value: Some("0"), subdenomination: None, kind: Number, argument_types: None, is_constant: false, is_l_value: false, is_pure: true, l_value_requested: false, type_descriptions: TypeDescriptions { type_identifier: Some("t_rational_0_by_1"), type_string: Some("int_const 0") }, src: "205:1:0", id: 6 }

.....
```

In the above sample, we can see that the `id` of `3` and `6` are repeated.

@alexroan This causes a problem when we do a traversal and encounter funny situations like 

```rust
if let X = node.parent {
    // but here `X`'s children doesn't contain `node`
}
```

Link to full output [here](https://app.warp.dev/block/Ryuiwkn0ZOF9chSoi8BeNZ) 

## My thesis [ yet to test ]

Since we can see above that `id` of `3` is repeating even within the same source index of `0` (`src: "128:3:0"` and `"166:1:0"`), what really makes the node unique is a combination of `NodeContext` and the current `node.id`.

### Follow up thoughts ...

When we have a `referenced_declarations` array, that is a vector of `NodeID`s what do we assume then? These `NodeID`s  share the same `NodeContext`  as the node that's hosting these properties ? If not, how do we uniquely distinguish them from the lot?

------------

### This PR

Working on a potential fix 
(atm I have made some mistakes but will test out the above thesis this weekend hopefully  🤞)

( I have hope it will be backward compatible thanks to `capture!` if successful)


alexroan:
Spitballing without digging too far into this yet:

Could this be to do with the fact that Foundry outputs multiple json files, one for each solidity version, for some contracts?

TilakMaddy:
@alexroan I agree that could be a possibility ... hence I included `if existing.src() != Some(&node.src)` This will hunt for nodes that appear at different offsets (roughly you can say different lines) in the *.sol file...

TilakMaddy:
```bash

Existing ID : 3
("src/ConstantsLiterals.sol", 6, "128:3")
Replacing with ID : 3
("src/cloc/HeavilyCommentedContract.sol", 14, "166:1")

Existing ID : 6
("src/ConstantsLiterals.sol", 7, "180:42")
Replacing with ID : 6
("src/cloc/HeavilyCommentedContract.sol", 16, "205:1")

Existing ID : 9
("src/ConstantsLiterals.sol", 8, "271:66")
Replacing with ID : 9
("src/cloc/HeavilyCommentedContract.sol", 19, "276:1")

Existing ID : 12
("src/ConstantsLiterals.sol", 10, "369:0")
Replacing with ID : 12
("src/cloc/HeavilyCommentedContract.sol", 22, "382:1")

Existing ID : 15
("src/ConstantsLiterals.sol", 11, "404:3")
Replacing with ID : 15
("src/cloc/HeavilyCommentedContract.sol", 29, "492:1")

Existing ID : 24
("src/ConstantsLiterals.sol", 13, "450:25")
Replacing with ID : 24
("src/cloc/HeavilyCommentedContract.sol", 35, "621:1")

Existing ID : 41
("src/ConstantsLiterals.sol", 18, "690:29")
Replacing with ID : 41
("src/cloc/HeavilyCommentedContract.sol", 40, "784:34")

Existing ID : 49
("src/ConstantsLiterals.sol", 20, "814:29")
Replacing with ID : 49
("src/cloc/HeavilyCommentedContract.sol", 40, "908:2")

Existing ID : 53
("src/ConstantsLiterals.sol", 10, "369:500")
Replacing with ID : 53
("src/cloc/HeavilyCommentedContract.sol", 40, "931:24")

Existing ID : 61
("src/ConstantsLiterals.sol", 26, "971:18")
Replacing with ID : 61
("src/cloc/HeavilyCommentedContract.sol", 43, "1007:1")

Existing ID : 69
("src/ConstantsLiterals.sol", 28, "1046:19")
Replacing with ID : 69
("src/cloc/HeavilyCommentedContract.sol", 47, "1130:41")

Existing ID : 73
("src/ConstantsLiterals.sol", 28, "1046:30")
Replacing with ID : 73
("src/cloc/HeavilyCommentedContract.sol", 49, "1268:34")

Existing ID : 77
("src/ConstantsLiterals.sol", 29, "1117:42")
Replacing with ID : 77
("src/cloc/HeavilyCommentedContract.sol", 49, "1341:1")

Existing ID : 56
("src/cloc/HeavilyCommentedContract.sol", 42, "965:12")
Replacing with ID : 56
("src/eth2/DepositContract.sol", 59, "4641:2")

Existing ID : 59
("src/ConstantsLiterals.sol", 25, "958:3")
Replacing with ID : 59
("src/eth2/DepositContract.sol", 61, "4752:1")
```

TilakMaddy:
This template can help re generate this locally

```rust
 fn visit_literal(&mut self, node: &Literal) -> Result<bool> {
        let x = self.nodes.insert(node.id, ASTNode::Literal(node.clone()));
        self.literals_context.insert(
            node.clone(),
            NodeContext {
                source_unit_id: self.last_source_unit_id,
                contract_definition_id: self.last_contract_definition_id,
                function_definition_id: self.last_function_definition_id,
                modifier_definition_id: self.last_modifier_definition_id,
            },
        );
        if x.is_some() {
            let existing = x.unwrap();
            // if existing.src() != Some(&node.src) {
            println!(
                "Existing ID : {}\n{:?}",
                &existing.id().unwrap(),
                self.get_node_sort_key(&existing)
            );
            println!(
                "Replacing with ID : {}\n{:?}\n",
                node.id,
                self.get_node_sort_key(&node.into())
            );
            // }
        }
        Ok(true)
    }
```

TilakMaddy:
Okay looks like I don't have a complete solution because even with my assumptions on talking the whole node context into picture, it becomes hard to distinguish what nodes we are talking about in case of 

```rust
#[derive(Clone, Debug, Deserialize, Eq, Serialize)]
#[serde(rename_all = "camelCase")]
pub struct Identifier {
    pub argument_types: Option<Vec<TypeDescriptions>>,
    pub name: String,
@>    pub overloaded_declarations: Vec<NodeID>,
@>    pub referenced_declaration: NodeID,
    pub type_descriptions: TypeDescriptions,
    pub src: String,
    pub id: NodeID,
}
```

TilakMaddy:
@alexroan I think I found the root cause ... But need your help in deciding where to go from here.

Step 1 -  if we do `forge build --ast --force` it will remove cache, out and rebuild the contract-playground

Step 2 - Now we will only find collisions of `NodeID`s between `HeavilyCommentedContract` and `DepositContract.sol`  (using above technique)  I think this is because the pragma versions for solidity in both these files were mutually exclusive.

DepositContract - `pragma solidity 0.6.11;`

HevailyCommentedContract - `pragma solidity >=0.4.22 <0.6.0;`

So when I bump up the upper limit of `HeavilyCommentedContract` to `pragma solidity >=0.4.22 <0.8.0;` all the NodeID collisions disappear.

Why do you think this happens? I'm unable to wrap my head around it. If forge is okay with building all these files in one project, why does `solc` exercise the right to reuse `NodeID`s when pragma versions are mutually exclusive ?

TilakMaddy:
@alexroan  My observation has been this - When we mutually exclusive pragma ranges in the contract playground - solc reserves its right to re-use NodeID for nodes across these mutually exclusive contract pairs. However, if you bump up the upper version of (say `HeavilyCommentedContract.sol` to `0.8.0` instead of `0.6.0` so it no longer is mutually exclusive to `DepositContract`) these collisions of IDs from nodes of different source units seem to disappear. 

So I think this is a new constraint in the system. We must warn users in the project if there are mutually exclusive pragma ranges : )

In fact we can discover such a case by using the above template and then say "Aderyn might not work properly given the project's circumstances" 

NOTE: You will still have some collisions of IDs but they would literally the same node being "visited" twice by WokspaceContext. But we can ignore these : )

I guess generally speaking though (at least in @Equious words' in most functional projects you wouldn't find version wise mutually exclusive contract pairs) we may not run into this issue in well maintained projects... 

(My pain point was the funny parent situation that I described above -I guess should be fixed in https://github.com/Cyfrin/aderyn/pull/220 )

TilakMaddy:
By the way `forge build --ast --force` is required when you make changes in pragma version and test. (in my experience)
Otherwise it just picks up from the same cached json files and it may waste a lot your time : )

alexroan:
Thank you for the research @TilakMaddy , this is great context.

I'm going to close this but open an issue with the discussion.

PR #221, opened 2024-03-17
Fix: Aderyn panics when empty sol file found in playground

TilakMaddy:
😅

When there is an empty file, `cloc.rs` panicked as follows!

```bash
Foundry src path: "./tests/contract-playground/src"
thread '<unnamed>' panicked at aderyn_core/src/fscloc/util.rs:22:27:
attempt to subtract with overflow
```

`cloc.rs` did not handle empty file cases. Now it does : )


aderyn_core/src/fscloc/cloc.rs
```diff
@@ -33,6 +33,10 @@ pub struct TokenInsightGroup {
 }
 
 pub fn get_stats(r_content: &str) -> Stats {
+    if r_content.is_empty() {
+        return Stats { code: 0 };
+    }
+
```
TilakMaddy, line 39: fixed here

alexroan (approved):

PR #222, opened 2024-03-17
`capture!` now handles ASTNode as well !

TilakMaddy:
What this PR does ?

* Allows `capture!` now handles ASTNode in addition to everything we have so far.
* NOT a breaking change 

See below for examples (as usual)

How it does it ?

1. Introducing `Capturable` enum that sees `ASTNode` on the *same* level as it sees other explicit node types.
2. That, combined with Rust's algebraic data types specifically `into()`  allow `capture!()` to create the tuple `(filename, line no, chopped loc)` without breaking change. (Also thanks to `WorkspaceContext` methods for hosting `Capturable` type in their formal arguments which means detector files do not need to import `Capturable` to use it : )


aderyn_core/src/detect/experimental/immediate_parent.rs
```diff
@@ -31,15 +31,15 @@ impl IssueDetector for ImmediateParentDemonstrator {
             if let Some(first_parent) = assignment.parent(context) {
                 if let ASTNode::Block(block) = first_parent {
                     println!("1 {}", block);
-                    capture!(self, context, block);
+                    capture!(self, context, first_parent);
```
TilakMaddy, line 34: Here, you can have `capture!(self, context, block)` as well as `capture!(self, context, first_parent)`

aderyn_core/src/detect/experimental/parent_chain.rs
```diff
@@ -30,11 +30,11 @@ impl IssueDetector for ParentChainDemonstrator {
...
             if let Some(parent_chain) = assignment.parent_chain(context) {
-                if let ASTNode::Block(block) = parent_chain[0] {
-                    capture!(self, context, block);
+                if let ASTNode::Block(_) = parent_chain[0] {
+                    capture!(self, context, parent_chain[0]);
                 }
-                if let ASTNode::ForStatement(for_statement) = parent_chain[1] {
-                    capture!(self, context, for_statement);
+                if let ASTNode::ForStatement(_) = parent_chain[1] {
+                    capture!(self, context, parent_chain[1]);
```
TilakMaddy, line 37: Here we do not have to bother deriving `for_statement`. We just capture it in the `ASTNode` form : )

aderyn_core/src/detect/experimental/parent_chain.rs
```diff
@@ -30,11 +30,11 @@ impl IssueDetector for ParentChainDemonstrator {
...
                 }
                 if let ASTNode::Block(block) = parent_chain[2] {
                     capture!(self, context, block);
```
TilakMaddy, line 40: If you want to derive the underlying type, you can still do it !

aderyn_core/src/context/capturable.rs
TilakMaddy, line 1: This file does the magic : )

aderyn_core/src/context/capturable.rs
```diff
@@ -0,0 +1,666 @@
...
+use super::workspace_context::{ASTNode, WorkspaceContext};
+
+#[derive(Clone)]
+pub enum Capturable {
+    ASTNode(ASTNode),
+    Assignment(Assignment),
+    BinaryOperation(BinaryOperation),
+    Block(Block),
+    Conditional(Conditional),
+    ContractDefinition(contracts::ContractDefinition),
```
TilakMaddy, line 12: ASTNode and the actual nodes are "seen" as though they are at the same level

aderyn_core/src/detect/experimental/immediate_parent.rs
alexroan: this is EPIC dude. Love it!

PR #223, opened 2024-03-17
Smart extractor

TilakMaddy:
In one traversal, extract multiple types of nodes from a given node's subtree

This is in contrast to calling a bunch of `ExtractXXX` methods from the same node.


TilakMaddy:
Closing because I don't think we have a need for this *now* .

PR #228, opened 2024-03-18
Cleanup: Remove aderyn_nodejs until solution found.

alexroan:
Cleaning the repo by removing aderyn_nodejs: This doesn't quite work for the reasons described [here](https://github.com/neon-bindings/neon/issues/117) (re: Neon).


alexroan:
Any pushback against this one @TilakMaddy ?

TilakMaddy:
That's alright @alexroan  : ) We'll figure something out later. 

My only comment would probably be that `aderyn_py` is not in as bad a shape as `aderyn_nodejs` (Given that the python bindings do work as expected unlike nodejs neon).

I would prefer to keep it and maintain it but ultimately I'd much rather defer the decision to you : )

TilakMaddy (approved):

alexroan:
@TilakMaddy good point, let's keep python bindings and iterate on them a little bit so they're fully functional.

PR #229, opened 2024-03-18
Added closest parent method on ASTNode

TilakMaddy:
If `x` is an `ASTNode`

Now it's possible to do the following

* `x.closest_parent_of_type(context, NodeType::Block)`

Other 

* code improvements - moved the `parent`, and `parent_chain` methods from `workspace_context.rs` to browser modules. 
(So that it's tied to a `GetImmediateParent`, `GetParentChain` traits and as a result it's consistent with how we call on explicit nodes)


aderyn_core/src/detect/experimental/immediate_parent.rs
```diff
@@ -32,7 +32,7 @@ impl IssueDetector for ImmediateParentDemonstrator {
                 if let ASTNode::Block(block) = first_parent {
                     println!("1 {}", block);
                     capture!(self, context, first_parent);
-                    if let Some(second_parent) = block.parent(context) {
+                    if let Some(second_parent) = first_parent.parent(context) {
```
TilakMaddy, line 35: example #1

aderyn_core/src/detect/experimental/closest_parent.rs
```diff
@@ -34,7 +34,7 @@ impl IssueDetector for ClosestParentDemonstrator {
                 capture!(self, context, block);
             }
 
-            if let Some(ASTNode::ForStatement(for_statement)) =
+            if let Some(for_statement) =
                 assignment.closest_parent_of_type(context, NodeType::ForStatement)
```
TilakMaddy, line 38: example #2

Down below we can see `for_statement.closest_parent_of_type(context, NodeType::Block)`  but this time `for_statement` is of type `&ASTNode`

alexroan (approved):

PR #237, opened 2024-03-19
Feature/ Implementation of `Node` interface for `ASTNode` unlocks the whole Extraction Library

TilakMaddy:
This PR implements Node for ASTNode.

What does it do?

* Allows for following patterns :

```rust
let x: ASTNode = block.parent(context); // or whichever way we get the ASTNode

// We can have 

ExtractAssignments::from(x).extracted;
ExtractBlocks:from(x).extracted; 
ExtractBinaryOperations::from(x).extracted;

```


alexroan (approved):
The abstractions just keep on coming from @TilakMaddy 😍

PR #246, opened 2024-03-24
Speed up cli/reportgen.sh by 10x

TilakMaddy:
Everytime we run `cli/reportgen.sh` we do not have to forge build the same `tests/contract-playground` 6 times for generating variations of reports.

This PR introduces an 'opt-in' env variable called `ADERYN_SKIP_BUILD` which skips that part of the pipeline.

So now, `cli/reportgen.sh` is lightning fast !


aderyn_core/src/framework/foundry.rs
```diff
@@ -65,14 +66,23 @@ pub fn load_foundry(foundry_root: &PathBuf) -> Result<LoadedFoundry, Box<dyn Err
...
+    let key = "ADERYN_SKIP_BUILD";
+
+    let should_build = match env::var(key) {
+        Ok(val) => val != "1",
+        Err(_) => true,
+    };
+
+    if should_build {
+        // Run `forge build --ast` in the root
+        let _output = std::process::Command::new("forge")
```
TilakMaddy, line 78: Here !

cli/reportgen.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+cd tests/contract-playground
+forge build --ast
+cd ../../
+
+export ADERYN_SKIP_BUILD=1
```
TilakMaddy, line 7: We build once at the beginning

cli/reportgen.sh
```diff
@@ -23,6 +29,7 @@ cargo run -- ./tests/contract-playground -o report.json
 # Create report-config.json based on config file
 cargo run -- --config-file ./tests/aderyn.config.json -o report-config.json ./tests/contract-playground/ 
 
+unset ADERYN_SKIP_BUILD=1
```
TilakMaddy, line 33: Then unset it : )

alexroan (approved):

PR #247, opened 2024-03-24
Utility function - immediate children on any Node (or) ASTNode

TilakMaddy:
.


TilakMaddy:
Again I'll merge the lint corrections from cross compile ci cd draft : )

aderyn_core/src/detect/experimental/immediate_children.rs
```diff
@@ -0,0 +1,98 @@
...
+            .filter(|x| x.name.contains("increment"))
+            .take(1)
+            .next()
+            .unwrap();
+
+        if let Some(children) = inc.body.as_ref().unwrap().immediate_children(context) {
+            for child in children {
+                assert!(
+                    child.node_type() == NodeType::IfStatement,
+                    "Only if statement should be caught in function body's immediate children!"
```
TilakMaddy, line 44: Here, `inc.body` is a function block:
That has a for loop and an if statement nested in it. However, when we call `immediate_children(context)` on the `&Block` node, we only get the for loop

aderyn_core/src/detect/experimental/immediate_children.rs
```diff
@@ -0,0 +1,98 @@
...
+        let mut detector = ImmediateChildrenDemonstrator::default();
+        let found = detector.detect(&context).unwrap();
+        assert!(found);
+
+        println!("{:?}", detector.instances());
+        println!(
+            "Total number of instances: {:?}",
+            detector.instances().len()
+        );
+        assert!(detector.instances().len() == 1);
```
TilakMaddy, line 96: Verified here. Only 1 ast node is caught and that's the for loop

TilakMaddy:
Contract we're talking about here is `ParentChainContract.sol`

aderyn_core/src/detect/experimental/immediate_children.rs
```diff
@@ -0,0 +1,98 @@
...
+            .function_definitions()
+            .into_iter()
+            .filter(|x| x.name.contains("increment"))
+            .take(1)
+            .next()
+            .unwrap();
+
+        if let Some(children) = inc.body.as_ref().unwrap().children(context) {
+            for child in children {
+                assert!(
```
TilakMaddy, line 42: I changed the method name to just `children()` I think it's more legible and intuitive (like you would start typing ".children" after the node)

alexroan (approved):

PR #248, opened 2024-03-24
Allow to find relative location between ASTNodes (`appears_after` and `appears_before`)

TilakMaddy:
I feel this will be helpful when writing a detector that can check for CEI (check -effects -interactions) patterns, or any other future patterns solidity maintainers endorse to prevent re-entrancy attacks (that are relatively location-dependent).


aderyn_core/src/detect/experimental/immediate_parent.rs
```diff
@@ -41,6 +41,19 @@ impl IssueDetector for ImmediateParentDemonstrator {
...
+                                    .unwrap());
+                                assert!(second_parent
+                                    .appears_after(context, third_parent)
+                                    .unwrap());
+                                assert!(second_parent
+                                    .appears_before(context, first_parent)
+                                    .unwrap());
+                                assert!(third_parent
+                                    .appears_before(context, second_parent)
+                                    .unwrap());
```
TilakMaddy, line 56: Given 2 `ASTNode`s,  you can query their relative location like this

aderyn_core/src/context/workspace_context.rs
```diff
@@ -1455,6 +1456,51 @@ impl WorkspaceContext {
...
+        match f.1.cmp(&s.1) {
+            Ordering::Less => Some(Ordering::Less),
+            Ordering::Equal => {
+                // If the nodes are on the same line, we must compare offset in the chopped_location
+                let first_character_offset = f.2.split_once(':').unwrap();
+                let second_character_offset = s.2.split_once(':').unwrap();
+                Some(first_character_offset.0.cmp(second_character_offset.0))
+            }
+            Ordering::Greater => Some(Ordering::Greater),
+        }
```
TilakMaddy, line 1482: Heart of the logic

TilakMaddy:
If this gets approved, I could work on making these methods `appears_before` and `appears_after` available on any general node, without having to convert to `ASTNode`

aderyn_core/src/context/workspace_context.rs
alexroan: Ahhhhh so it uses the src location

alexroan (approved):
Love this.

We should maintain a list of these helpers somewhere in the `detect` directory so that contributors are aware of them.

TilakMaddy:
But yeah maybe like these lower level utilities in the `detect` folder....

alexroan:
I mean in a README so that contributors know what is available to them when they're writing detectors :)

PR #253, opened 2024-03-27
Enhancements/make abstractions scalable

TilakMaddy:
The problem currently is when you introduce an abstraction you have to implement it on all ~40 AST nodes  and if the parameters allow for an AST Node number of methods you add exponentially increases

so if you want to have a function

`f(x, y)`

You need to write 40 * 40  functions of you want to access x.id and y.id

We want to reduce it without having to force  call `into()` on the node and making it an `ASTNode`

Solution

This PR  allows allow you to achieve your goal with just writing **1** function

--------

Examples have been implemented in all browser modules so now the rust compiler does the creation of these methods for static dispatch - we no longer have to implement these browser methods on each node

Note: NO BREAKING Changes

--------

How this was achieved?

By allowing access to `id` on anything that implements `Node` interface.


aderyn_core/src/context/browser/location.rs
TilakMaddy, line 1: This file has a new design pattern to write abstractions (Write once, run for any node :P )

aderyn_core/src/detect/experimental/immediate_parent.rs
```diff
@@ -45,6 +45,10 @@ impl IssueDetector for ImmediateParentDemonstrator {
                                 assert!(first_parent
                                     .appears_after(context, second_parent)
                                     .unwrap());
+                                assert!(first_parent
+                                    .appears_after(context, for_statement)
+                                    .unwrap());
+                                assert!(block.appears_after(context, for_statement).unwrap());
```
TilakMaddy, line 51: So now `appears_after` and `appears_before` works on any general node and even AST Node (or combo)

TilakMaddy:
AA actually i want to try to solve this in a *slightly* different way... But the same concept tho : ) that's why I'm drafting it

aderyn_core/src/visitor/ast_visitor.rs
```diff
@@ -497,15 +497,26 @@ pub trait ASTConstVisitor {
...
+    /// [`Node::accept_metadata`] is designed to propagate into the AST subtree
+    /// although it doesn't happen by itself. [`Node::accept`] triggers the propagation
     fn accept_metadata(&self, _visitor: &mut impl ASTConstVisitor) -> Result<()> {
         Ok(())
     }
+    /// [`Node::accept_id`] is not designed to propagate into the AST subtree
+    fn accept_id(&self, _visitor: &mut impl ASTConstVisitor) -> Result<()> {
+        Ok(())
+    }
 }
```
TilakMaddy, line 520: @alexroan  Here I have documented the fact that `accept_id()` is not propagated (so there is no front processing)

TilakMaddy: Speed tradeoff in terms of compilation time may increase tho : ) Cuz the compiler has to create these methods  for all types that implement Node :P

alexroan: You mean compilation of `aderyn` tho, right? That's completely and utterly fine :D

alexroan (approved):

PR #255, opened 2024-03-28
New detector: create2 instead of create

TilakMaddy:
Create2 opcode makes contract deployment less susceptible to re-orgs and also you can predict the adress so  we will have much more consistent deployments.

Example taken from Solidity by example

https://solidity-by-example.org/new-contract/


tests/contract-playground/src/Create.sol
```diff
@@ -0,0 +1,77 @@
...
+        cars.push(car);
+    }
+
+     function createAndSendEther2(address _owner, string memory _model)
+        public
+        payable
+    {
+        // BAD
+        Car car = new Car{value: msg.value}(_owner, _model);
+        cars.push(car);
```
TilakMaddy, line 42: Create2 is not used here. So we capture them

tests/contract-playground/src/Create.sol
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    function getCar(uint256 _index)
+        public
+        view
+        returns (
+            address owner,
+            string memory model,
+            address carAddr,
```
TilakMaddy, line 69: Create2 is used here. So we don't capture them

PR #256, opened 2024-03-28
New detector: Library should not have public or external functions

TilakMaddy:
It's not ideal to have public or external functions in libraries as they are not designed to interact with other external contracts and or maintain state. Hence it can be misleading.


tests/contract-playground/src/LibraryContract.sol
```diff
@@ -0,0 +1,32 @@
...
+                x = (y / x + x) / 2;
+            }
+        } else if (y != 0) {
+            z = 1;
+        }
+        // else z = 0 (default value)
+    }
+    // BAD
+    function remove(uint256[] storage arr, uint256 index) public {
+        // Move the last element into the place to delete
```
TilakMaddy, line 21: Example here

aderyn_core/src/detect/nc/library_has_public_external_funcs.rs
```diff
@@ -0,0 +1,81 @@
...
+            .filter(|x| x.kind == ContractKind::Library)
+        {
+            for function in contract.function_definitions() {
+                if function.visibility == Visibility::Public
+                    || function.visibility == Visibility::External
+                {
+                    capture!(self, context, function);
+                }
+            }
+        }
```
TilakMaddy, line 31: Logic here

alexroan:
I'm not convinced this is worthy of being something that Aderyn recommends as a security concern right now. Let's shelf this.

TilakMaddy:
Got it : ) Turns out I have a reference for this one - 

(LightChaser's findings for GOAT.TECH Cantina competition - https://cantina.xyz/competitions/f214cf86-cc80-40c0-a70b-e9bb25d7ac80)

But can I conclude that  Aderyn's flavor differs slightly from what we see in bot reports for competitions?

PR #257, opened 2024-03-28
New Detector: Unprotected initialize function

TilakMaddy:
Detects unprotected `initialize` function. Without any modifiers/safety guards initialize can be susceptible to frontrunning attacks

So the idea is that when the function name contains `init` there needs to be 

a) either a modifier to that function (which we can assume will take care of access rights and revert when necessary)

b) or "revert" / "require" statements somewhere in the function

If the above is not satisfied it's a bad initializer function so we capture it : )


tests/contract-playground/src/UnprotectedInitialize.sol
```diff
@@ -0,0 +1,39 @@
...
+        require(!initialized, "Contract already initialized");
+        initialized = true;
+        // Additional initialization logic here
+    }
+
+    // BAD
+    function initializeWithoutModifierOrRevert() external {
+        initialized = true;
+        // Additional initialization logic here
+    }
```
TilakMaddy, line 38: Contract File here

aderyn_core/src/detect/low/unprotected_init_function.rs
```diff
@@ -0,0 +1,77 @@
...
+                        .iter()
+                        .any(|x| x.name == "revert" || x.name == "require")
+                    {
+                        capture!(self, context, function);
+                    }
+                }
+            }
+        }
+
+        Ok(!self.found_instances.is_empty())
```
TilakMaddy, line 34: logic here

alexroan (approved):

PR #259, opened 2024-03-28
New Detector : Global time units can be used instead of custom definitions

TilakMaddy:
.


aderyn_core/src/detect/nc/time_units_not_used.rs
```diff
@@ -0,0 +1,77 @@
...
+    // Keys are source file name and line number
+    found_instances: BTreeMap<(String, usize, String), NodeID>,
+}
+
+impl IssueDetector for TimeUnitsNotUsedDetector {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
+        let disallow = [
+            "second", "seconds", "minute", "minutes", "hour", "hours", "day", "days", "week",
+            "weeks", "month", "months",
+        ];
```
TilakMaddy, line 22: When we see variables names that match with this list, we recommend using natively available units like `1 days` , etc in solidity

alexroan: `month` isn't a time unit in Solidity

aderyn_core/src/detect/nc/time_units_not_used.rs
```diff
@@ -0,0 +1,77 @@
...
+
+impl IssueDetector for TimeUnitsNotUsedDetector {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
+        let disallow = [
+            "second", "seconds", "minute", "minutes", "hour", "hours", "day", "days", "week",
+            "weeks",
+        ];
+
+        for variable in context.variable_declarations() {
+            if disallow.iter().any(|&x| x == variable.name.to_lowercase()) {
```
alexroan, line 25: This feels too broad to me, and at the same time wouldn't catch many cases. This would catch:

```solidity
uint256 private constant DAY = 1 days;
```

but wouldn't catch:

```solidity
uint256 private constant 2_DAYS = 60*60*24*2;
```

alexroan (changes requested):

alexroan:
I've added some extra examples in the Solidity test contract. I think we should rethink this one @TilakMaddy , IMO it should be based on the right hand side of the assignment, not the name on the left. Closing for now.

PR #261, opened 2024-03-29
Feature `peek` into node to see the source code as string + `TODO` Detector 

TilakMaddy:


aderyn_core/src/context/browser/peek.rs
```diff
@@ -0,0 +1,12 @@
...
+pub trait Peek {
+    /// Peek into the source code of a node
+    fn peek(&self, context: &WorkspaceContext) -> Option<String>;
+}
+
+impl Peek for ASTNode {
+    fn peek(&self, context: &WorkspaceContext) -> Option<String> {
+        context.get_source_code_of_node(self.id()?)
+    }
+}
```
TilakMaddy, line 12: Right now, it only works on ASTNode but if my other PR https://github.com/Cyfrin/aderyn/pull/253 is merged then I can easily extend this to all nodes natively

aderyn_core/src/context/workspace_context.rs
```diff
@@ -1231,6 +1231,28 @@ impl WorkspaceContext {
...
+        let src_location = node.src().unwrap_or("");
+
+        let chopped_location = match src_location.rfind(':') {
+            Some(index) => &src_location[..index],
+            None => src_location, // No colon found, return the original string
+        }
+        .to_string();
+
+        if let Some((offset, len)) = chopped_location.split_once(':') {
+            let offset: usize = offset.parse().ok()?;
```
TilakMaddy, line 1246: peek logic here

aderyn_core/src/detect/nc/contracts_with_todos.rs
```diff
@@ -0,0 +1,91 @@
...
+    fscloc,
+};
+use eyre::Result;
+
+#[derive(Default)]
+pub struct ContractsWithTodosDetector {
+    // Keys are source file name and line number
+    found_instances: BTreeMap<(String, usize, String), NodeID>,
+}
+
```
TilakMaddy, line 20: Contract with todo Detector here

tests/contract-playground/src/ContractWithTodo.sol
```diff
@@ -0,0 +1,17 @@
...
+    
+    // Function to add a new task
+    function createTask(string memory _content) private {
+        // TODO: Add task to the tasks array
+        // TODO: Emit TaskCreated event
+    }
+
+    // Function to toggle the completion status of a task
+    function toggleCompleted(uint _taskId) public {
+        // TODO: Toggle the completed status of the task
```
TilakMaddy, line 14: Sample todo here

aderyn_core/src/context/browser/peek.rs
TilakMaddy: > Right now, it only works on ASTNode but if my other PR #253 is merged then I can easily extend this to all nodes natively

DONE ✅  (now every `impl Node` has a `peek()`)

alexroan:
Waiting on Benchmarking

TilakMaddy:
Alex, I should have added a benchmarking script for an individual detector. I think you didn't run it here because the current script does so for all of them and it's time-consuming. Sorry 😞

PR #268, opened 2024-03-30
Detector fix/enhancement : Do not catch internal functions used only once that start with `_`

TilakMaddy:


aderyn_core/src/detect/nc/useless_internal_function.rs
```diff
@@ -22,7 +22,10 @@ impl IssueDetector for UselessInternalFunctionDetector {
         let internal_functions = context
             .function_definitions()
             .into_iter()
-            .filter(|&function| matches!(function.visibility, Visibility::Internal));
+            .filter(|&function| {
+                matches!(function.visibility, Visibility::Internal)
+                    && !function.name.starts_with('_')
+            });
```
TilakMaddy, line 29: Main change (that's all)

... everything else is reportgen

tests/contract-playground/src/InternalFunctions.sol
```diff
@@ -23,4 +24,9 @@ contract InternalFunctionExample {
         // This function is called only once within the contract.
         value = _newValue;
     }
+
+    function internalSet2(uint256 _newValue) internal {
+        // This function is called only once within the contract.
+        value = _newValue;
+    }
```
TilakMaddy, line 31: Added this function.

TilakMaddy: It doesn't start with `_` so we capture it. Now, there is another function called `_internalSet` - although it's used only once we don't capture it because the programmer has acknowledged it by starting the function name with an `_`

TilakMaddy:
I got this idea after generating [this](https://gist.github.com/TilakMaddy/bf21b4de205195f6606d4e8dd4f4fa23) report for chainlink ccip

alexroan (approved):

PR #269, opened 2024-03-30
Force build forge in `cli/reportgen` so that `out` folder doesn't stale

TilakMaddy:
.


TilakMaddy:
Could this be a caching error in github runner?

cli/reportgen.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 cd tests/contract-playground
-forge build --ast
+forge build --ast --force
```
TilakMaddy, line 4: Changing this forces forge to build AST without using cache files in `out` folder and this reduces the risk of MEP-related bugs  🐞  in generating report. (Helps in long term and healthy maintenance of this codebase)

However, we don't update the command within `aderyn` because we prioritize speed there.

alexroan:
@TilakMaddy I'm seeing very different `out` files that basically undo all the `out` changes you have here, demonstrated in this PR: #272 .

forge --version: `forge 0.2.0 (d94e3c6 2024-04-01T00:19:02.563573000Z)`

TilakMaddy:
> @TilakMaddy I'm seeing very different `out` files that basically undo all the `out` changes you have here, demonstrated in this PR: #272 .
> 
> forge --version: `forge 0.2.0 (d94e3c6 2024-04-01T00:19:02.563573000Z)`

Hmm that strange . I have to look into it - while i was expecting slightly different outfiles cause of some randomness factor , i was NOT expecting tests to fail. I'll look into it

alexroan:
It could be to do with the solidity compiler. We remove metadata in the `foundry.toml` because of it uses local machine data to append a hash to bytecode. There could be something else there that we're not accounting for.

I'm going to close this and my referenced PR for now. Let's reopen at a later date and figure this out.

Issue #264, opened 2024-03-29
Doesn't respect Foundry profiles


RensR:
We use many different Foundry profiles and it would be nice if Aderyn could respect the active Foundry profile to scope to the configured folder.

Tested using aderyn 0.0.18

```
➜  contracts git:(golf) ✗ export FOUNDRY_PROFILE=ccip
➜  contracts git:(golf) ✗ aderyn                     
Detecting framework...
Framework detected: Foundry mode engaged.
Foundry root path: "."
error: unexpected argument '--ast' found

Usage: forge build [OPTIONS]

For more information, try '--help'.
Foundry src path: "./src/v0.8"
```

The src path of the ccip profile should be "./src/v0.8/ccip" not the generic root path displayed here.

TilakMaddy:
Please run `foundryup`

RensR:
```
➜  contracts git:(golf) ✗ foundryup
foundryup: installing foundry (version nightly, tag nightly-617dfc28cb8206a0003edcf73a6f1058adaef740)
foundryup: downloading latest forge and cast
######################################################################################################################################################################################################################################################### 100.0%
foundryup: downloading manpages
######################################################################################################################################################################################################################################################### 100.0%
foundryup: installed - forge 0.2.0 (617dfc2 2024-03-29T00:21:19.015083000Z)
foundryup: installed - cast 0.2.0 (617dfc2 2024-03-29T00:21:18.899529000Z)
foundryup: done
➜  contracts git:(golf) ✗ aderyn   
Detecting framework...
Framework detected: Foundry mode engaged.
Foundry root path: "."
[⠆] Compiling...
No files changed, compilation skipped
Foundry src path: "./src/v0.8"
Error reading Solidity source file: "src/v0.8/ccip/test/mocks/MockBridgeAdapter.sol"
No such file or directory (os error 2)
```

Foundry src path is still "./src/v0.8" and it is still processing files outside of the `ccip` path which is the profile that is active

TilakMaddy:
Can you share foundry.toml (if you don't mind) ?

TilakMaddy:
AHH Got it !! Yes, as of now `aderyn` only looks at the default profile. We need to fix that : ) 

Thanks @RensR

TilakMaddy:
Culprit https://github.com/Cyfrin/aderyn/blob/6cd765556f95c271ea938e0b9fce8780de25b489/aderyn_core/src/framework/foundry.rs#L27

PR #270, opened 2024-03-30
Accommodate for non default `FOUNDRY_PROFILE` 

TilakMaddy:
Fix https://github.com/Cyfrin/aderyn/issues/264


TilakMaddy:
Will add some tests and then undraft it : )

TilakMaddy:
The fix is ready... but there's a problem in how we do our  verification steps in CI pipeline  (we have to account for undeterministic errors due to NodeID collisions):P I'll work on that first

aderyn_core/src/framework/foundry.rs
```diff
@@ -113,23 +98,85 @@ pub fn load_foundry(foundry_root: &PathBuf) -> Result<LoadedFoundry, Box<dyn Err
...
+        let default_profile = foundry_profiles.get("default");
+
+        for (profile_key, value) in foundry_profiles {
+            if profile_key.contains(&foundry_profile) {
+                if let Some(value) = value.as_table() {
+                    let profile_src = {
+                        if let Some(toml::Value::String(src)) = value.get("src") {
+                            src.clone()
+                        } else if let Some(default_profile) = default_profile {
+                            if let Some(toml::Value::String(src)) = default_profile.get("src") {
```
TilakMaddy, line 136: If `src` is not found in the `foundry.toml` section of the relevant profile, we fallback to the default profile

Similarly for `out`

tests/contract-playground/foundry.toml
```diff
@@ -2,4 +2,7 @@
 src = "src"
 out = "out"
 libs = ["lib"]
-bytecode_hash = "none"
\ No newline at end of file
+bytecode_hash = "none"
+
+[profile.sample]
+out = "sample_out"
```
TilakMaddy, line 8: `sample` profile

report.sample_profile.md
TilakMaddy, line 1: This is the report generated when using `FOUNDRY_PROFILE=sample`

cli/reportgen.sh
```diff
@@ -11,6 +12,9 @@ export ADERYN_SKIP_BUILD=1
 # Basic report.md 
 cargo run -- ./tests/contract-playground &
 
+# Basic report.profile.md 
+FOUNDRY_SAMPLE=sample cargo run -- -o report.sample_profile.md ./tests/contract-playground &
+
 # Create report-config.md based on config file
```
TilakMaddy, line 18: Automatically generate report for the `sample` FOUNDRY_PROFILE. 

Note: At the moment it may be same in content as `report.md` but it need not be in the future as we reserve the right to change the `src` for the `sample` profile in `foundy.toml` later sometime. 

Let's not link both the reports : )

aderyn_core/src/framework/foundry.rs
```diff
@@ -15,34 +16,13 @@ pub struct FoundryOutput {
...
-
-#[derive(Debug, Deserialize)]
-struct DefaultProfile {
-    #[serde(default = "default_src")]
     src: String,
-    #[serde(default = "default_out")]
     out: String,
 }
 
-fn default_src() -> String {
```
alexroan, line 38: These should definitely not be removed!

alexroan (changes requested):
This currently breaks the current foundry behavior.

For example, you could have a foundry.toml that doesn't define `src` and `out`, so the [defaults are used](https://book.getfoundry.sh/reference/config/project#src).

This PR doesn't currently account for that, and therefore aderyn doesn't run fully over a perfectly valid repo that it previously did.

TilakMaddy:
Noted. I understood now, so there is defaults even for the default profile  : ) 

I'll correct it by evening!

TilakMaddy:
Summary of changes

First correction: 

Previous behavior
Look for `$FOUNDRY_PROFILE` in foundry.toml and in that section  look for `src` and `out` in default profile. (and if the default profile doesn't explicitly define `src` and `out`, throw an error)

Corrected behavior
Look for `$FOUNDRY_PROFILE` in foundry.toml and in that section  look for `src` and `out` in environment variables  as defined [here](https://book.getfoundry.sh/reference/config/project#src) and then have a fallback of literal string `"src"` and `"out"`

aderyn_core/src/framework/foundry.rs
```diff
@@ -113,23 +98,76 @@ pub fn load_foundry(foundry_root: &PathBuf) -> Result<LoadedFoundry, Box<dyn Err
...
+                            default_foundry_src
+                        }
+                    };
+
+                    let profile_out = {
+                        if let Some(toml::Value::String(out)) = value.get("out") {
+                            out.clone()
+                        } else {
+                            default_foundry_out
+                        }
```
TilakMaddy, line 152: Now it's a lot cleaner here  : )

aderyn_core/src/framework/foundry.rs
```diff
@@ -113,23 +98,76 @@ pub fn load_foundry(foundry_root: &PathBuf) -> Result<LoadedFoundry, Box<dyn Err
...
+        .to_lowercase();
+
+    // Official Foundry docs defaults
+    // https://book.getfoundry.sh/reference/config/project#src
+
+    let default_foundry_src = std::env::var("FOUNDRY_SRC")
+        .unwrap_or(std::env::var("DAPP_SRC").unwrap_or(String::from("src")));
+
+    let default_foundry_out = std::env::var("FOUNDRY_OUT")
+        .unwrap_or(std::env::var("DAPP_OUT").unwrap_or(String::from("out")));
```
TilakMaddy, line 131: Default values are preserved here

TilakMaddy:
Summary of changes

New behavior
Look inside `$FOUNDRY_PROFILE` (if not set, default to `"default"`) section in `foundry.toml`  for `src` and `out` values

If they don't exist, the look  in environment variables  as defined [here](https://book.getfoundry.sh/reference/config/project#src) and also give them a  fallback of literal string `"src"` and `"out"`

alexroan (approved):

PR #281, opened 2024-04-05
New Detector: weak randomness

TilakMaddy:
Capture the following patterns for weak randomness and suggest Chainlink VRF

1. `uint256(keccak256(...))`
2. `uint(keccak256(..))`


TilakMaddy:
@alexroan I could complete this but I feel like it's a good starter idea for hackathon. The goal is simple : ) I'm adding this as an issue with hackathon tag as well

PR #286, opened 2024-04-05
fail on forge build --ast

TilakMaddy:
Related to issue https://github.com/Cyfrin/aderyn/issues/285


TilakMaddy:
This is quite surprising :P

TilakMaddy:
I think the problem is, in CI we are trying to execute `forge build --ast` in a directory that does not exist (invalid foundry_absolute_path)

Issue #69, opened 2023-12-08
Formal release process (CI/CD?)


alexroan:
Publishing versions of:
- crates
- python bindings
- node bindings

alexroan:
Hey @TilakMaddy , I've read through the doc. Most makes sense.

What would be the best next steps here? My instinct is to focus on Python bindings first: Create bash scripts that can be used to publish the bindings, and test those out.

TilakMaddy:
Yes Alex...I will address your questions shortly

lukapodlesnik:
@alexroan  you should add a docker image to the list. When an ew version is released CI/CD automatically builds the container image from GitHub and push it Docker Hub.
If you are planing to use Github actions this would be the flow: https://docs.github.com/en/actions/publishing-packages/publishing-docker-images

alexroan:
I'm restricting the scope of this to formal CI releases, which is captured in #514

PR #290, opened 2024-04-07
Benchmarks for Aderyn and individual detectors 

TilakMaddy:
Run `cargo bench` to see output like [this](https://app.warp.dev/block/ttZlrYP2pcKSPHAR7d9Kfk)

* every time you run the command it creates and stores reports in `target/criterion` (HTML files, charts, etc)
* in the CLI output itself you can see which detectors have increased in perf and which ones have regressed
* once you delete `target/criterion` you lose history


aderyn_driver/Cargo.toml
```diff
@@ -10,4 +10,11 @@ license = "MIT"
 
 [dependencies]
 aderyn_core = { path  = "../aderyn_core", version = "0.0.19" }
-rayon = "1.8.0"
\ No newline at end of file
+rayon = "1.8.0"
+
+[dev-dependencies]
+criterion = "0.5.1"
+
```
TilakMaddy, line 17: This is the crate added as a dev dependency that helps us in benchmarking

TilakMaddy:
@alexroan  Shall we add benchmarking template in nyth?

TilakMaddy:
Draft because:

I want to add a benchmark for the overall aderyn process (where all detectors run in parallel),

not just for individual detectors sequentially (what we currently have)

alexroan:
> I want to add a benchmark for the overall aderyn process (where all detectors run in parallel),
> 
> not just for individual detectors sequentially (what we currently have)

*Both* would be epic @TilakMaddy 👍

alexroan:
> @alexroan Shall we add benchmarking template in nyth?

Not right now. I'm considering whether we actually need `Nyth` at all tbh.

alexroan:
@TilakMaddy Do you still have TODOs on this one?

TilakMaddy:
@alexroan I have added `cli/benchmarks.sh`. So now it does a couple of things

1) Copy the existing benchmarks (which I am attaching with this PR) into `target/`
2) Run the bench
3) Update the bench reports here (move them from `target/` to `benchmarks`)
4) Open the report in the browser to the main page where you can click on any of the options ( individual detector name (or) aderyn) performance and see how it has regressed/improved based on prev. No relying on CLI output

TilakMaddy:
I'd prefer if CI could do this for every release and automatically commit the new reports to the codebase : )

alexroan (approved):
This scaffolding suffices for now. We'll do some work after this PR as part of #69 to add this to release flow.

PR #291, opened 2024-04-07
New browser module: previous and next sibling

TilakMaddy:


aderyn_core/src/detect/experimental/sibling.rs
```diff
@@ -0,0 +1,73 @@
...
+impl IssueDetector for SiblingDemonstrator {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
+        for contract in context.contract_definitions() {
+            if let Some(children) = contract.children(context) {
+                if let Some(sorted) = children.sort_by_line_nos(context) {
+                    assert!(sorted.len() >= 2);
+                    assert!(sorted[1].previous_sibling(context).unwrap() == sorted[0]);
+                    assert!(sorted[0].next_sibling(context).unwrap() == sorted[1]);
+                }
+            }
```
TilakMaddy, line 30: In practice, these methods can act as shortcuts for traversal.

aderyn_core/src/context/browser/siblings.rs
```diff
@@ -0,0 +1,76 @@
...
+
+        // Find the ID of the node this method is called upon
+        self.accept_id(&mut node_id_receiver).ok()?;
+        let current_node_id = node_id_receiver.id?;
+
+        let parent = context.get_parent(current_node_id)?;
+        let children = parent.children(context)?;
+        let sorted_children = children.sort_by_line_nos(context)?;
+
+        for i in 0..sorted_children.len() - 1 {
```
alexroan, line 45: Is there a more efficient way of doing this loop?

The cost varies based on the number of siblings, and the position of the current sibling. Is there a way we can make this cost the same no matter the position&sibling numbers?

aderyn_core/src/context/browser/siblings.rs
```diff
@@ -0,0 +1,76 @@
...
+        let mut node_id_receiver = NodeIDReceiver::default();
+
+        // Find the ID of the node this method is called upon
+        self.accept_id(&mut node_id_receiver).ok()?;
+        let current_node_id = node_id_receiver.id?;
+
+        let parent = context.get_parent(current_node_id)?;
+        let children = parent.children(context)?;
+        let sorted_children = children.sort_by_src_position(context)?;
+
```
TilakMaddy, line 44: This is `O(n log n)`  which is sorting the array.

TilakMaddy: This dominates in time over  the below `O(n)` loop

aderyn_core/src/context/browser/siblings.rs
TilakMaddy: I'll check if the IDs in are also sorted when nodes are sorted by src position. If yes, we could do binary search

PR #292, opened 2024-04-08
Fix/forge not installed in ci

TilakMaddy:


aderyn_core/src/framework/foundry.rs
```diff
@@ -60,13 +60,19 @@ pub fn load_foundry(foundry_root: &PathBuf) -> Result<LoadedFoundry, Box<dyn Err
...
             .stderr(Stdio::inherit())
-            .status();
+            .status()
+            .expect("Failed to run `forge build --ast`");
+
+        if !output.success() {
+            eprintln!("The command `forge build --ast` did not execute successfully");
+            std::process::exit(output.code().unwrap());
+        }
     }
```
TilakMaddy, line 77: Now it passes : )

TilakMaddy:
I can't believe we came this far without having forge 😂🫣


👽

aderyn_core/src/framework/foundry.rs
```diff
@@ -60,13 +60,19 @@ pub fn load_foundry(foundry_root: &PathBuf) -> Result<LoadedFoundry, Box<dyn Err
...
-        let _output = std::process::Command::new("forge")
+        let output = std::process::Command::new("forge")
             .arg("build")
             .arg("--ast")
             .current_dir(&foundry_root_absolute)
             .stdout(Stdio::inherit()) // This will stream the stdout
             .stderr(Stdio::inherit())
-            .status();
+            .status()
+            .expect("Failed to run `forge build --ast`");
```
alexroan, line 70: This may not be the only error. It may be that the contracts fail to compile also. We should account for all errors in this command.

alexroan (changes requested):

aderyn_core/src/framework/foundry.rs
TilakMaddy: @alexroan But we already handle that by checking if the exit code is success or not in the next few lines

aderyn_core/src/framework/foundry.rs
TilakMaddy: If the contracts fail to compile, output.success() will be false

aderyn_core/src/framework/foundry.rs
TilakMaddy: This particular line only panics if forge is not available as a command in the env

alexroan: You are correct, apologies!

alexroan (approved):

PR #297, opened 2024-04-09
Integration/ Foundry-Compilers

TilakMaddy:
`load_contract_directly('path/to/solidity_file')`  

* Doesn't depend on `forge` availability in the path
* Doesn't create any `out/*` files
* Doesn't create any virtual workspace


aderyn_driver/src/driver.rs
```diff
@@ -205,3 +205,131 @@ fn detect_framework(path: PathBuf) -> Option<Framework> {
...
+            .current_dir(tests_contract_playground_path)
+            .stdout(Stdio::piped())
+            .output()
+            .unwrap();
+
+        let stdout = String::from_utf8(command.stdout).unwrap();
+
+        println!("AST {}", stdout);
+        assert!(!stdout.is_empty());
+    }
```
TilakMaddy, line 334: We generate the AST with just `solc`  (also this solc version is detected based on the pragma content in the file, then auto installed) We manually run solc tho : )

TilakMaddy:
I'm positive on this <3 

Some stuff needs to be done. But overall it looks like a good idea : )

alexroan:
> Wanted to try out a quick experiment : )
> 
> Please let me know if running `cargo test foundry_compiler_tests -- --nocapture` in this branch outputs a bunch of AST on your screen : )

It does @TilakMaddy !

TilakMaddy:
UPDATE: While `--stop-after parsing` is a wonderful option, it WON'T WORK FOR US. We have to rely on full compilation only.

aderyn_driver/src/driver.rs
```diff
@@ -205,3 +205,204 @@ fn detect_framework(path: PathBuf) -> Option<Framework> {
...
+            // assert!(!stdout.is_empty());
+        }
+    }
+
+    #[test]
+    fn can_load_context_for_admin_contract() {
+        let cargo_root = PathBuf::from(std::env::var("CARGO_MANIFEST_DIR").unwrap());
+        let tests_contract_playground_path = cargo_root
+            .join("../tests/contract-playground/")
+            .canonicalize()
```
TilakMaddy, line 337: @alexroan My goal for this PR is pretty scoped down. I don't want to make the whole aderyn project independent of foundry/hardhat JUST YET.

I want to first test this out with our `load_contract()` function which is what we use when we write tests for our detectors.

So I mean `load_contract('/blah/../src/path/to/solidity/file.sol')` should automatically download the relevant solc for that file, create ast and prepare the context loader. (caveat: solc would be spawned from `src` directory. Because We don't have any configuration yet, let's stick to defaults)

Maybe a bit of observation on how it performs (gather feedback, etc), and then after that let's tackle the beast! What do you say?

TilakMaddy: Update: I'll use `load_contract_experimental` because I'm not even sure how to get these tests to pass in CI lol. I'm looking for some help in that aspect too :)

aderyn_core/src/detect/detector.rs
```diff
@@ -266,4 +273,93 @@ pub mod detector_test_helpers {
         });
         context
     }
+
+    pub fn load_contract_directly(filepath: &str) -> WorkspaceContext {
+        let solidity_file = &ensure_valid_solidity_file(filepath);
```
TilakMaddy, line 278: BOOM !

aderyn_core/src/detect/low/empty_blocks.rs
```diff
@@ -102,4 +105,28 @@ mod empty_block_tests {
             String::from("Consider removing empty blocks.")
         );
     }
+
+    #[test]
+    fn test_empty_block_by_loading_contract_directly() {
+        let context = load_contract_directly("../tests/contract-playground/src/EmptyBlocks.sol");
+
+        let mut detector = EmptyBlockDetector::default();
```
TilakMaddy, line 113: Here it detects  0.8.19 and uses the relevant solc compiler

aderyn_core/src/detect/experimental/sibling.rs
```diff
@@ -70,4 +73,13 @@ mod sibling_demo_tests {
         let mut detector = SiblingDemonstrator::default();
         let _ = detector.detect(&context).unwrap();
     }
+
+    #[test]
+    fn test_siblings_by_loading_contract_directly() {
+        let context =
+            load_contract_directly("../tests/contract-playground/src/StorageConditionals.sol");
+
+        let mut detector = SiblingDemonstrator::default();
```
TilakMaddy, line 82: Here it detects  0.8.20 and uses the relevant solc compiler

TilakMaddy:
@alexroan The tests should pass locally. Let me know please 🙏  Also help me with making it pass in CI

aderyn_core/Cargo.toml
```diff
@@ -20,6 +20,8 @@ serde_json = "1.0.96"
 serde_repr = "0.1.12"
 strum = { version = "0.26", features = ["derive"] }
 toml = "0.8.2"
+foundry-compilers = { git = "https://github.com/foundry-rs/compilers", features = ["svm-solc"] }
```
alexroan, line 23: We should lock this to a specific commit.

aderyn_core/src/detect/detector.rs
```diff
@@ -266,4 +273,91 @@ pub mod detector_test_helpers {
...
+        if let Ok(command) = command {
+            let stdout = String::from_utf8(command.stdout).unwrap();
+
+            let mut pick_next_line = false;
+            let mut ast_content = String::new();
+            for line in stdout.lines() {
+                if line.starts_with("======= ") {
+                    let end_marker = line.find(" =======").unwrap();
+                    let filepath = &line["======= ".len()..end_marker];
+                    if file_arg.contains(filepath) {
```
alexroan, line 316: This loop seems to exclude base contracts.

In the previous step, solc outputs AST for all inherited and target contracts. This line only picks the one that matches the filename, meaning that `ast_content` doesn't include the base contracts, and we don't get the inherited details of those base contracts.

aderyn_core/src/detect/detector.rs
```diff
@@ -266,4 +273,94 @@ pub mod detector_test_helpers {
...
+
+        let file_arg = compiler_input
+            .sources
+            .first_key_value()
+            .unwrap()
+            .0
+            .to_str()
+            .unwrap();
+
+        let command = Command::new(solc_bin)
```
alexroan, line 299: In CI, this is returning nothing for the EmptyBlocks test that's failing 🤔

alexroan:
I think the `solc` command is erroring in CI only, and I'm trying to get reasonable debug output for it, hence the many commits.

TilakMaddy:
Thanks a lot, @alexroan  for investigating so much : ) I'll see where I can take it from here. Still a bit confused why this doesn't happen locally but it's okay at least there is *something*  now - we know that the relevant solc binary gets downloaded successfully and has permission to execute.

I am going to read this fully - https://docs.soliditylang.org/en/v0.8.25/path-resolution.html

alexroan:
@TilakMaddy I'll also close this. Please continue working from the new base branch which #333 is based on! :)

PR #321, opened 2024-04-11
Staging v0.0.21

alexroan:
* Updates to Contributing and Issue guidelines
* Reduce the number of severity categories to just two: High and Low
  * Highs represent things that should definitely addressed
  * Low represent nice to haves, but not mandatory


TilakMaddy:
> * Updates to Contributing and Issue guidelines
> * Reduce the number of severity categories to just two: High and Low
>   * Highs represent things that should definitely addressed
>   * Low represent nice to haves, but not mandatory

What about the distinction between not following coding best practices (example event not emitted on state change) ... This doesn't harm the protocol 
...
And

Low vulnerabilities (ex - weak randomness function used)... Could harm the protocol

alexroan:
> > * Updates to Contributing and Issue guidelines
> > * Reduce the number of severity categories to just two: High and Low
> >   
> >   * Highs represent things that should definitely addressed
> >   * Low represent nice to haves, but not mandatory
> 
> What about the distinction between not following coding best practices (example event not emitted on state change) ... This doesn't harm the protocol ... And
> 
> Low vulnerabilities (ex - weak randomness function used)... Could harm the protocol
> 
> ...
> 
> So I'm thinking like - low, high, bestpractice, (or something like that)

@TilakMaddy We may evolve into that 3-tier separation depending on user feedback. We're starting with 2 for simplicity, so that it's very easy to write rules about what is High and what is Low, primarily for the hackathon. 

High = Something that should definitely be fixed, because it _could_ break your contracts
Low = Nice to haves - but not "must fix"

PR #327, opened 2024-04-13
BASE BRANCH: Aderyn Compilation Framework

TilakMaddy:
😉

Run the following and let me know if you see the `fcr.md`  with all the issues. If so, then we are on the verge of becoming independent of `forge` and `hardhat`

`cargo run -- ./tests/contract-playground/ --scope src/ --exclude lib/ -o fcr.md`

Note: Prereq https://github.com/Cyfrin/aderyn/pull/297


aderyn_driver/src/process_foundry.rs
TilakMaddy, line 1: Kill

aderyn_driver/src/process_hardhat.rs
TilakMaddy, line 1: Kill

aderyn_driver/src/virtual_foundry.rs
TilakMaddy, line 1: Kill

TilakMaddy:
When I remove the starting part of the markdown reports (file summary, file details, etc).... and i do a diff - I see that there is NO difference in issues reported in report.md (where we used forge) and fcr.md (where nothing was required - of course the relevant solc is installed and used on fly if not present)

TilakMaddy:
Arghh now i know why there is repetition in filenames for file summary and files details section - it's the same reason why forge creates multiple `.json` files for a given sol file : )

TilakMaddy:
UPDATE: We now get identical results with solc  !!!

report.md
```diff
@@ -69,7 +69,6 @@ This report was generated by [Aderyn](https://github.com/Cyfrin/aderyn), a stati
 | src/UnsafeERC721Mint.sol | 18 |
 | src/ZeroAddressCheck.sol | 41 |
 | src/cloc/AnotherHeavilyCommentedContract.sol | 32 |
-| src/cloc/EmptyContractFile.sol | 0 |
 | src/cloc/HeavilyCommentedContract.sol | 21 |
```
TilakMaddy, line 72: We don't show empty files any more (That's the only difference)

TilakMaddy:
@alexroan Could you try it out and review it already? These tests pass locally but not in CI having a hard time finding out why

cli/reportgen.sh
```diff
@@ -1,42 +1,30 @@
...
-FOUNDRY_PROFILE=sample cargo run -- -o report.sample_profile.md ./tests/contract-playground &
+cargo run --  --scope src/ --exclude lib/ ./tests/contract-playground &
 
 # Create report-config.md based on config file
-cargo run -- --config-file ./tests/aderyn.config.json -o report-config.md ./tests/contract-playground/  &
+cargo run -- --config-file ./tests/aderyn.config.json --exclude lib/ -o report-config.md ./tests/contract-playground/  &
 
 # Create report.judge.md 
-cargo run -- ./tests/contract-playground -o judgeops/current/report.judge.md &
+cargo run --  --scope src/ --exclude lib/ -o judgeops/current/report.judge.md  ./tests/contract-playground &
```
TilakMaddy, line 12: Although we don't rely on frameworks, I think we must take clues for scope and exclude from famous files `foundry.toml`

TilakMaddy:
First of all thanks for confirming tests and report generation.

My thoughts ...

1. Yeah I am most certain the test issue is the same as #297  (coz this #327 is branched out from #297)
2. Sure, merge conflict resolution can be done. I'll do it : )
3. That's a good point. I didn't think about it - all I thought was - we're only using stdout everything should be fast. But losing the ability to cache is where we could take a hit. However, if the cost we pay by recompiling ASTs for all files is less than Foundry's cost of generating bytecode, abi, etc in the JSON then I think it's still a better deal  (Just me spitballing) And we should use benchmarks to figure that out : ) If that's not the case then one the strategies I can think of is to build our own  aderyn-backend that does what foundry-backend does (which is create all these outfiles that can be used as cache) but do with our configuration. So that means we're doing `solc --ast-compact-json <file1> <file2>` and storing the output in out files and then returning cached output if the file content has not changed. Now that brings me to another question - how does the cost of parsing AST compare to reading a file. If it's not too much then maybe caching is not very helpful. Because if it takes the same time to read and determine if the file has changed as it does to parse the file into an AST string then we don't have to work so hard : P 

I agree on having a new base branch for this. It looks like it needs polishing

alexroan:
Closing this for renamed base branch: #333

TilakMaddy:
Sorry I clicked the wrong button

TilakMaddy:
Sorry again ...fat finger😭

PR #330, opened 2024-04-15
Remove ENV vars in favour of CLI args

### LINKED ISSUE #302: Align all ENV variables and CLI args

alexroan:
Resolves #302


alexroan:
@TilakMaddy I've removed all ENV-related stuff here and defaulted to CLI args instead. If we ever do need ENV vars also, we'll introduce them for every CLI arg.

PR #334, opened 2024-04-16
Remove openSSL dependency

alexroan:
This dependency was added in the update check feature, and blocks some users who don't have open SSL. We will add this back eventually, but it's not worth adding the dependency atm

Revert "Notify aderyn users when a new version is released ! (#263)"
This reverts commit c4570a76e24d082c01dff688268c296acc0cf09b.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,22 +3,11 @@
 version = 3
 
 [[package]]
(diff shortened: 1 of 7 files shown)


alexroan:
cc @TilakMaddy

PR #338, opened 2024-04-16
New abstractions: `peek_over` and `peek_under` (essential for some detectors)

TilakMaddy:
The existing `peek()` detector gives you the text *inside the node and bordering the node* .

This is not enough if we want to detect comments above function definitions (ex: you want to check for @notice, @dev comments, etc)

So this PR introduces 

* `peek_over()` This returns text outside the node up until the previous node (if it exists else it will stretch until the top of the source unit)
*  `peek_under()` This returns text outside up until the next node (if it exists else it will stretch until the bottom of the source unit)

With this PR, we can add the following  hackathon ideas in issues
Missing natspec for 
1) Contracts
2) Structs
3) Functions
4) Public variables


aderyn_core/src/detect/experimental/peek_over.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+impl IssueDetector for PeekOverDemonstrator {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
+        for event in context.event_definitions() {
+            if let Some(text) = event.peek_over(context) {
+                if text.contains("// SAME CONDITIONALS")
+                    || text.contains("// DIFFERENT CONDITIONALS")
+                {
+                    capture!(self, context, event);
```
TilakMaddy, line 24: Example here - we check for events that have these comments just above them

alexroan (approved):
Very cool, this will help us when we later implement something like:

```solidity

// @aderyn-ignore
uint256 something = reallyBad.happened();

```

PR #339, opened 2024-04-17
Staging v0.0.24

alexroan:
* Reintroduce update checking but without using OpenSSL in the request crate
* Slither acknowledgment in the Readme
* New AST traversal abstractions: `peek_over` and `peek_under`
* Update License to GNU GPL
* https://docs.cyfrin.io/ added to readme
* `load_multiple_contracts` test helper
* New detector: low, reverts and requires inside loops


alexroan:
@TilakMaddy we're getting errors when Checking reports, specifically with the FOUNDRY_PROFILE run.

Not sure what's causing them atm because I can't reproduce locally. Can you take a look?

TilakMaddy:
> @TilakMaddy we're getting errors when Checking reports, specifically with the FOUNDRY_PROFILE run.
> 
> UPDATE 1: Previously successful CI runs (like on the `master` branch after the last release) are also now failing when I rerun them 🤔 It might be a race condition with one of the detectors?
> 
> I can't reproduce locally. Can you take a look?
> 
> UPDATE 2: More investigation and context in #341
> 
> UPDATE 3: Resolved in #341 :)

I'm seeing this just now. Sorry

PR #341, opened 2024-04-17
CI Check Reports Debugging

alexroan:
Screenshot 2024-04-17 at 15 12 47
Screenshot 2024-04-17 at 15 13 13

Differences between:
* Locally generated version (left)
* CI generated (right)

Both concern the `ContractWithTodos.sol`


alexroan:
AAAHHHH There might have been some IDs conflicting in older compiled versions of some contracts in `sample_out`. Removed them and this seems to have solved the issue.

PR #346, opened 2024-04-18
Detector: Low, unused imports

alexroan:


aderyn_core/src/detect/low/unused_imports.rs
```diff
@@ -0,0 +1,202 @@
...
+            });
+        }
+        None => {
+            distant_relatives.push(-99);
+        }
+    }
+}
+
+impl IssueDetector for UnusedImportsDetector {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
```
alexroan, line 47: @TilakMaddy Can you review this and let me know what you think? I've added some long comments to try to describe what's going on here.

TilakMaddy: Surely will look into it : )

TilakMaddy:
Question: What would you do differently if we had a way to load the dependencies' ASTs? (coz its totally possible with our new internal compilation framework)

TilakMaddy:
I'm a little confused in the first test case - So what I see is if I remove the `import ./Counter.sol` line, it still compiles. This means Counter is an unused import. So why do we not capture that ?

TilakMaddy:
Pattern that we can use instead of creating a new function 

```rust
if let Some(node) = context.nodes.get(xxx.referenced_declaration.id) {
   node.closest_ancestor_of_type(context, NodeType::SourceUnit)
}
```

alexroan:
> Question: What would you do differently if we had a way to load the dependencies' ASTs? (coz its totally possible with our new internal compilation framework)

@TilakMaddy this was my offline monologue while thinking about this. It should give you more context of where I'm at:

### Current Problem
The report is printing way more false positives than expected when running `cargo run` , but not in tests. This is because in the tests, we're explicitly loading all of the imported files, as well as the source files. During a normal run, only JSON output files matching the files in `src` are being loaded in.

### Possible Solutions - Brainstorm
1. Load all JSON, but only load the source files in `src`
	1. All source_units within scope (`src`) would have their source lines loaded into memory, but imports would not. This is OK in terms of not breaking the existing detectors, but it will slow aderyn down.
		1. To resolve this, especially for when we have a vscode extension, in single-file-mode we might need more information from the output json file to load in all the exact imported json files? This should work, but involves some parsing before visiting each AST node.
		2. Question: How would this work with --scope and --exclude?
			1. It should be fine - the main problem will be that when you then use _any_ getter in the WorkspaceContext, it will return imported source_units and other types from the imported files. All ASTNodes from imported files. This could be fine? It will be another performance hit though, especially if there are a bunch of imported contracts - however, foundry will only compile the ones that are actually used... but also there are hangover output files that were used in previous compilations, and left hanging in newer compilations if they were removed from the imports of contracts in the source folder. These may have stale IDs that overlap with newer compilations. FK.

### Further Thoughts
The premise of the detector itself is sound, the problem lies with the way in which aderyn loads in the contracts and ASTs initially. If we can fix that, without fucking up other detector results, then we're golden.

We don't want to load in _all_ AST output files, because of the lingering overlap problem. However, unless we introduce a system which loads specific AST JSON files during the initial loading, we have no way of implementing this detector successfully - by that, meaning that it accounts for imports _outside_ of the foundry source folder.

Let's try to implement just that for now - with a comment for this limitation with libs.

I have implemented this basic form of detector, but it does not work with imports that have a lineage outside of the scope. Further work needed.

alexroan:
Closing due to staleness, will revisit later.

PR #353, opened 2024-04-21
New Detector:  reverts and requires inside for loops / while loops

TilakMaddy:
Avoid `require` / `revert` statements in a loop because a single bad item can cause the whole transaction to fail. It's better to forgive on fail and return failed elements post-processing of the loop


TilakMaddy:
It has been on my mind for a while now....though I'll knock it out today : P

tests/contract-playground/src/RevertsAndRequriesInLoops.sol
```diff
@@ -0,0 +1,35 @@
...
+        }
+    }
+
+    function deductMassFundsV2() external {
+        for (uint256 id = 0; id < 10; ++id) {
+            if(funds[id] == 0) {
+                revert();
+            }
+            delete funds[id];
+        }
```
TilakMaddy, line 22: Bad

tests/contract-playground/src/RevertsAndRequriesInLoops.sol
```diff
@@ -0,0 +1,35 @@
...
+     function deductMassFundsV3() external returns(uint256[] memory failedIds) {
+        uint256[] memory failedIds = new uint256[](10);
+        for (uint256 id = 0; id < 10; ++id) {
+            if(funds[id] == 0) {
+                failedIds[id] = 1;
+            }
+            delete funds[id];
+        }
+    }
+
```
TilakMaddy, line 34: Good

aderyn_core/src/detect/low/reverts_and_requries_in_loops.rs
```diff
@@ -0,0 +1,103 @@
...
+impl IssueDetector for RevertsAndRequiresInLoopsDetector {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
+        // Collect all require statements
+        let requires_and_reverts = context
+            .identifiers()
+            .into_iter()
+            .filter(|&id| id.name == "revert" || id.name == "require")
+            .collect::<Vec<_>>();
+
+        for item in requires_and_reverts {
```
alexroan, line 30: @TilakMaddy an interesting future benchmark will be to test this retrieval method, vs:
1. Get all loops
2. Extract reverts and requires from them

alexroan (approved):
Looks good! Merging :)

PR #355, opened 2024-04-22
Support for different styles of `load_contract_directly`

TilakMaddy:
`load_source_units()` maybe a more accurate thing to say

### DIFF
--- a/aderyn_core/src/detect/detector.rs
+++ b/aderyn_core/src/detect/detector.rs
@@ -358,6 +358,60 @@ pub mod detector_test_helpers {
         }
     }
 
+    pub struct ContractLoader {
+        compiler_input: CompilerInput,
+    }
+
+    impl ContractLoader {
+        pub fn lock() -> impl Drop {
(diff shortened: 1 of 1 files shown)


aderyn_core/src/detect/detector.rs
```diff
@@ -358,6 +358,60 @@ pub mod detector_test_helpers {
...
+
+    pub enum ContractLoaderInput {
+        SingleSourceUnit(String),
+        SourceUnitWithAllDependencies(String),
+        SourceUnitWithAllDependenciesScopedAndExcluded(
+            String,
+            Option<Vec<String>>,
+            Option<Vec<String>>,
+        ),
+    }
```
TilakMaddy, line 380: 3 ways of supplying input for `load_contract_directly`

TilakMaddy: (WIP)

TilakMaddy:
I liked the idea. tested it out. but I think we can park it here. revisit it after we convert all the detectors tests to loading contracts directly

PR #333, opened 2024-04-16
BASE BRANCH: Internal compilation framework

alexroan:
Resolves #298 

Tasks:
* Remove dependency on Foundry or Hardhat being installed locally.  ✅ 
* Remove dependency on the workspace being a Foundry or Hardhat project.  ✅ 
* Do not require any prerequisite steps like installing `solc` locally before running aderyn. ✅ 
* **Maintain feature parity with current version of aderyn**
* Version lock foundry-compilers dependency  ✅ 
* Handle Remappings   ✅ 
* ~Explore caching mechanisms (we no longer can take forge's help)~ (After merge to dev)
* Add `--icf` flag to opt-in ✅ 

Completed Feature parity
* Markdown report ✅
* JSON report ✅
* Sarif report ✅
* Auditor mode ✅
* CI tests for all reports (both ICF and legacy) ✅ 

Feature parity caveats
* Works on foundry project as-is ✅

* Works on hardhat project 
a) In ICF mode must pass  `--src contracts/` and have `remappings.txt` in project root) ✅
b) In legacy mode (a.k.a without the `--icf` flag) ther is NO support.  ✅

* Single file mode - TODO

Tentative (✅)

~Browse commonly known files and environment vars like `foundry.toml` if present,  to find out scope (by looking at src attribute value in foundry profile)~ We use foundry's official config crate to achieve this

Breaking changes  
* `requirements` or `requirements.txt` needs to be present in the project root if needed (not needed for foundry)

New changes
*  Introducing `--src` which is a comma-separated list of paths relative to the project root. Aderyn will only look inside the sources and follow `--scope` and `--exclude` (if set)

* This `--src` file only works in `--icf` mode

PR #358, opened 2024-04-22
Separation for detector test utils (new module)

TilakMaddy:
* Existing `load_contract` and `load_contract_directly` are now in 2 separate files under the same module (for breathing space)

* `load_contract_directly` has been renamed to `load_solidity_source_unit`


aderyn_core/src/detect/experimental/sibling.rs
```diff
@@ -78,9 +78,10 @@ mod sibling_demo_tests {
     }
 
     #[test]
+    #[serial(fc_solc)]
```
TilakMaddy, line 81: Tests with key `fc_solc` wrapped in `#[serial()]` will happen in sequential order. This was done to avoid the solc binary poisoned/busy error

TilakMaddy: More context about error here https://github.com/Cyfrin/aderyn/pull/333#issuecomment-2068039301

TilakMaddy:
If this passes, I'd like to merge it into the internal-compilation-framework branch

PR #359, opened 2024-04-22
Added direct contract loading for more detectors

TilakMaddy:
(Just playing around following the advice given by Dani Popes here  https://github.com/foundry-rs/compilers/issues/113#issuecomment-2069159149 )


TilakMaddy:
Multiple detector tests works !!!

(serial execution tho)

PR #360, opened 2024-04-22
Convert all remaining detector tests to loading contracts directly

TilakMaddy:


TilakMaddy:
Any thoughts?

The error we see in CI is `Error running solc command`

PR #368, opened 2024-04-25
New Detector: Constant with same name has multiple definitions 

TilakMaddy:
Constants with the same variable name declared  (state vars only) multiple times in different locations. 

This may lead to an exploit. It is always safe to have 1 source of truth in like an abstract contract where the reused constants are defined once for all the contracts to use


aderyn_core/src/detect/low/const_multiple_definitions.rs
```diff
@@ -0,0 +1,114 @@
...
+        low::const_multiple_definitions::ConstantDefinedMultipleTimesDetector,
+    };
+
+    #[test]
+    fn test_constants_with_multiple_definitions() {
+        let context = load_multiple_contracts(vec![
+            "../tests/contract-playground/out/C1.sol/C1.json",
+            "../tests/contract-playground/out/C2.sol/C2.json",
+        ]);
+
```
TilakMaddy, line 91: Playing around with the new load multiple contracts function

aderyn_core/src/detect/low/const_multiple_definitions.rs
```diff
@@ -0,0 +1,114 @@
...
+                    e.insert(vec![var_decl.to_owned()]);
+                }
+            }
+        }
+
+        for (_, const_potentially_defined_multiple_times) in vars_by_name.into_iter() {
+            let mut locations: HashSet<_> = HashSet::new();
+            for c in &const_potentially_defined_multiple_times {
+                locations.insert(context.get_node_sort_key(&c.into()));
+            }
```
TilakMaddy, line 46: I find that in the current setup, there can be duplicate nodes (2 different node IDs pointing to the same node)

That's why had to do this.

TilakMaddy:
@alexroan I'm unable to regen some JSON files in `out/` : (

After I ran `cli/reportgen.sh` some old files seem to have disappeared. I never cleared any cache. 

I could fix it in the PR but would be a pain point for hackathon

aderyn_core/src/detect/low/const_multiple_definitions.rs
TilakMaddy: @alexroan I think this is because foundry's out files have ASTs of the same contract which are coming from different compilation groups

tests/contract-playground/src/multiple_constants/C1.sol
```diff
@@ -0,0 +1,5 @@
+pragma solidity 0.8.20;
+
+contract C1 {
+    uint256 constant public MC = 123;
+}
```
TilakMaddy, line 5: MC variable here

tests/contract-playground/src/multiple_constants/C2.sol
```diff
@@ -0,0 +1,5 @@
+pragma solidity 0.8.20;
+
+contract C2 {
+    uint256 constant public MC = 456;
+}
```
TilakMaddy, line 5: Also here

Issue #369, opened 2024-04-25
Bug: Index out of bound panic in `get_source_code_of_node`


jf-li00:
**Describe the bug**
When I run the example mentioned in #301 , the program panicked due to an index out of bound error.

```
thread '<unnamed>' panicked at aderyn_core/src/context/workspace_context.rs:1254:48:
byte index 9636 is out of bounds of `// SPDX-License-Identifier: MIT
```

in `release` mode it is another contract but the same problem happens:
```
thread '<unnamed>' panicked at aderyn_core/src/context/workspace_context.rs:1254:48:
byte index 8801 is out of bounds of `// SPDX-License-Identifier: MIT
```

**To Reproduce**
1. run on this repo: https://github.com/smartcontractkit/ccip/tree/ccip-develop/contracts 
2. See error

jf-li00:
I spent some time debugging and found that before the painc happens, in the function `get_source_code_of_node` in `aderyn_core/src/context/workspace_context.rs`,  
the `node_id` variable is `159001` . How ever, the `id` field of `node` variable and `source_unit` variable  are both `159002`.
Since the `node` variable is declared as `let node = self.nodes.get(&node_id)?;`, I think `node.id == node_id` should hold

TilakMaddy:
Thanks for bringing this up!

I'll send a fix

TilakMaddy:
@jf-li00 Can you check it once after the PR gets in and let us know?

jf-li00:
@TilakMaddy I've checked it, this time it won't panic like what I mentioned before :)

jf-li00:
But we still can't find the right content of the source due to the duplication of AST node ids in `WorkspaceContext` introduced bt `solc`

TilakMaddy:
Yes you are right, my PR only fixes the pipeline from panicking. In reality, the node id replacement problem still exists. But we plan on fixing that in our new internal compilation framework that we are working on.

In that, we have the ability to break down different contexts while ensuring there is no repeating node IDs. Also the best part is we no longer rely on foundry toolchain installation, hardhat, etc.

The framework will download the relevant solc binaries and create multiple `WorkspaceContext`s for each project. (And as far as writing the detector goes, nothing changes. Only the internals from our end change) 

I'll keep you posted : )

alexroan:
[...]

PR #371, opened 2024-04-25
Fix: peek panics

TilakMaddy:
Sometimes for node properties `offset`, `len` can have values like `-1` 

That's why it's important we check the bounds before we return the string

Fix for https://github.com/Cyfrin/aderyn/issues/369

### DIFF
--- a/aderyn_core/src/context/browser/peek_over.rs
+++ b/aderyn_core/src/context/browser/peek_over.rs
@@ -39,12 +39,19 @@ impl<T: Node + ?Sized> PeekOver for T {
         if let Some(previous_sibling) = self.previous_sibling(context) {
             let (prev_offset, prev_len) =
                 context.get_offset_and_length_of_node(previous_sibling.id()?)?;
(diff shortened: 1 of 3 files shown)


alexroan:
Waiting for Hackathon submissions to conclude before addressing this PR 👍

alexroan (changes requested):
Could you add tests to check for these negative values please @TilakMaddy ?

Any time we fix something that's breaking, we should introduce tests at the same time :)

TilakMaddy:
I made a mistake. I don't think it's because of negative values (because we are deserializing it to `usize`). It has to be something else. Let me re run this without the checks on https://github.com/smartcontractkit/ccip/tree/ccip-develop/contracts to find out what values they were (which causes index out of bounds panic)

alexroan:
@TilakMaddy is this still relevant ?

TilakMaddy:
> @TilakMaddy is this still relevant ?

Yes Alex, we should get this PR in because it's important to check that index lies within bounds before accessing array element at that index

PR #375, opened 2024-04-27
Adhoc sol files example playground added (relative imports only)

TilakMaddy:
I have added a folder with a bunch of sol files and relative imports. That's it (no foundry.toml, hardhat.config.ts, etc)


TilakMaddy:
Follow up:
Same thing, but with direct imports (so for that we would need remappings)

More about direct imports vs relative imports [here](https://docs.soliditylang.org/en/v0.8.23/path-resolution.html)

tldr;
Direct imports are subjected to remappings first, whereas relative imports are directly given to HostFileSystemLoader 
Relative imports are imports that start with `./` or `../`

PR #381, opened 2024-04-29
Aderyn watch

TilakMaddy:
Check out this PR and try running the following command:

`cargo run -- ./tests/contract-playground --scope src/ --exclude lib/ -o report.json --watch --stdout`

It will constantly output JSON report to stdout whenever a file inside the project root is added, deleted, or modified.

You can remove `--stdout` to read directly from a live `report.json`

Note: debounce time is 2 seconds here. So if there are too many file changes, it will wait to observe a time period of at least 2 seconds where nothing changes, and then the detectors and the reporter kicks in


aderyn/src/main.rs
```diff
@@ -204,14 +217,84 @@ fn main() {
...
                     stdout: args.stdout,
                 };
-                driver::drive_with(new_args, subscriptions);
+                if cmd_args.watch {
+                    println!("INFO: Aderyn is entering watch mode !");
+                    // setup debouncer
+                    let (tx, rx) = std::sync::mpsc::channel();
+
+                    // no specific tickrate, max debounce time 2 seconds
+                    let mut debouncer = new_debouncer(Duration::from_secs(2), None, tx).unwrap();
```
alexroan, line 226: Changing this value doesn't seem to change the refresh rate.
Values used: 0, 2, 5. The refresh duration _seems_ the same.

Does debounce time only kick in when many files are being changed, or just one?
It could be that the delay is just the compilation framework and aderyn running.

alexroan (approved):
This is great dude, far less code than I expected.

aderyn/src/main.rs
TilakMaddy: In my understanding, debounce kicks in when it observes a period of 2 seconds when there are no file changes irrespective of how many they are. (1 or bulk)

TilakMaddy: But I need to look more into since you said the refresh rate doesn't seem to change

TilakMaddy: Plan for now is to schedule an interval of 2 seconds then read from report.json when we see event  `vscode/textDocumentSaved` (I don't remember the name exactly)

TilakMaddy: There are 2 places where we declare 2 seconds. Did you try changing in both places?

aderyn/src/main.rs
```diff
@@ -204,14 +217,84 @@ fn main() {
...
+
+        // Then run only if file change events are observed
+        if cmd_args.watch {
+            println!("INFO: Aderyn is entering watch mode !");
+            // setup debouncer
+            let (tx, rx) = std::sync::mpsc::channel();
+
+            // no specific tickrate, max debounce time 2 seconds
+            let mut debouncer = new_debouncer(Duration::from_secs(2), None, tx).unwrap();
+
```
TilakMaddy, line 279: Change here as well @alexroan : )

Issue #316, opened 2024-04-11
High Detector: OverUnderflow for solidity versions <0.8


alexroan:
For contracts that can be compiled using solidity versions below 0.8, if mathematical operations exist (`x + y`, for example), capture the operation and recommend that SafeMath is used.

PR #386, opened 2024-04-30
feat: Underflow/Overflow Detector for Solidity compiler versions <0.8

gr4yha7:
ref #316.

Solodit Issue: https://solodit.xyz/issues/unsafe-math-calculations-halborn-seascape-ninja-spin-pdf


report.md
```diff
@@ -296,6 +298,78 @@ Consider protecting the initializer functions with modifiers.
...
+- Found in src/Counter.sol [Line: 23](tests/contract-playground/src/Counter.sol#L23)
+
+	```solidity
+	        number += 2;
+	```
+
+- Found in src/Counter.sol [Line: 29](tests/contract-playground/src/Counter.sol#L29)
+
+	```solidity
+	        number += TWO;
```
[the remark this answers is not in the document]
gr4yha7: It's a bit weird because the detector's unit test doesn't find such instances when ran with those contracts. I'll find a way to fix it.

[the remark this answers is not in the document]
gr4yha7: It didn't work out. False positives are still in the report.

aderyn_core/src/detect/high/arithmetic_underflow_overflow.rs
```diff
@@ -0,0 +1,162 @@
...
+        assert_eq!(
+            detector.title(),
+            String::from("`SafeMath` library should be imported and used to prevent underflow and overflow in solidity compilers below `0.8.0`")
+        );
+        // assert that the detector returns the correct description
+        assert_eq!(detector.description(), String::from("Use `SafeMath` library for arithmetic operations to avoid underflow and/or overflow in solidity compilers below `0.8.0`"));
+    }
+
+    #[test]
+    fn test_arithmetic_underflow_overflow_constants_file() {
```
alexroan, line 138: @gr4yha7 I added this test to check a file that shouldn't capture any instances, but it does. (Please pull this branch)

It looks like your detector may still be catching contracts that have a pragma >=0.8.0.

alexroan (changes requested):

aderyn_core/src/detect/high/arithmetic_underflow_overflow.rs
```diff
@@ -0,0 +1,136 @@
...
+            _ => {}
+        }
+    }
+
+    false
+}
+
+impl IssueDetector for ArithmeticUnderflowOverflowDetector {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
+        for pragma_directive in context.pragma_directives() {
```
alexroan, line 50: I see the problem. Here, you're retrieving all pragma directives, but later on line 67, you're retrieving all assignments separately. There's no link between the pragmas that you've checked and the assignments that you retrieve from the WorkspaceContext. Therefore, the pragma check becomes redundant.

To resolve:
1. Get all `source_units`
2. Find the pragma -> do the check. If the pragma is below 0.8.0...
3. Extract all the assignments from the source unit
4. Capture the unsafe math

alexroan (changes requested):

aderyn_core/src/detect/high/arithmetic_underflow_overflow.rs
gr4yha7: @alexroan Right indeed! Thanks! 🚀

tests/contract-playground/src/ArithmeticUnderflowOverflow.sol
```diff
@@ -0,0 +1,122 @@
+// SPDX-License-Identifier: MIT
+pragma solidity <0.8.0;
+ 
+contract ArithmeticUnderflowOverflow {
+    uint256 public counter;
+
+    // bad
```
alexroan, line 7: I've added a lot more test cases to this file. The functions commented `// bad` should be caught by this detector, and the ones commented `// good` should not be.

tests/contract-playground/src/ArithmeticUnderflowOverflow.sol
```diff
@@ -0,0 +1,122 @@
...
+    function incrementCounter2() public {
+      counter++;
+    }
+
+    // bad
+    function decrementCounter2() public {
+      counter--;
+    }
+
+    // good
```
alexroan, line 47: This currently is captured by the detector, even though it is a false-positive.

This function is `// good` because there is a check within the function before the arithmetic operation occurs to ensure no underflow happens. This is what SafeMath does under the hood.

There is a way we can alter this detector so that it doesn't capture this as a false-positive. The trick is to look for:
* arithmetic operations on identifiers (which the detector almost does already, it just doesn't track the identifiers yet)
* other checks that occur within the same function, on the same identifiers.

If both are true, then the detector probably shouldn't capture this.

aderyn_core/src/detect/high/arithmetic_underflow_overflow.rs
```diff
@@ -0,0 +1,144 @@
...
+
+                let req = VersionReq::parse(&version_string)?;
+                if version_req_below_0_8(&req) {
+                    let assignments = ExtractAssignments::from(source_unit).extracted;
+                    let binary_operations = ExtractBinaryOperations::from(source_unit).extracted;
+
+                    assignments
+                        .iter()
+                        .filter(|assignment| {
+                            assignment.operator == "+=" || assignment.operator == "-="
```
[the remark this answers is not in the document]
alexroan: Also multiplication and division?

alexroan (changes requested):
Hey @gr4yha7, thank you so much for working on this, this so happens to be a detector that we should go deeper on! Always a fun process!
You've stumbled on a very interesting problem regarding false-positives.  

But before we go into the details, a quick word on this PR with regards to the Hackathon...

____

I'm going to conclude the hackathon winners PR #388 (which you already have a detector merged for), but keep this one open so we can continue to improve it before merging it. There's no time limit on this one. As long as you resolve all my reviews and it gets merged eventually, it will still be considered for a hackathon prize at a later date. :)

____

Let's talk about the details:
* I've added more cases to your test smart contract `ArithmeticUnderflowOverflow.sol`. These cases include `require` and `revert` conditionals prior to and after the arithmetic operations, similar to how SafeMath does it. These test cases should not be captured because the checks adequately deal with the under and overflow cases.
* I've also added functions using `++` and `--` which aren't captured yet.
* I've annotated `// bad` and `// good` above each of these functions to indicate which should be captured and not captured, respectively.

The question is, "how can we remove the false positives?". 

In the `// good` example test cases I've added, you'll notice there are conditionals and require statements that also use the same Identifiers as the arithmetic operations in the same function. The trick is to find the patterns between these, and when there is a match, don't capture it. Does that make sense?

I've dropped a few more review comments to help you. Please don't hesitate to ask questions, as me and @TilakMaddy will try to help as much as we can :)

tests/contract-playground/src/ArithmeticUnderflowOverflow.sol
gr4yha7: @alexroan Got it. I'll tackle these.

alexroan:
Closing due to staleness

PR #391, opened 2024-05-01
[Testing ICF] Fails on CCIP contracts. Help needed

TilakMaddy:
I think that [ccip](https://github.com/smartcontractkit/ccip/tree/ccip-develop/contracts) repo is a very good master tester of internal communication framework. 

We must absolutely make sure it works on this repo!

Currently, it fails with little to no text in `stderr`!  
Help needed debugging 🙏

 
> ....
> stderr = "" is not found.
>  ....
> thread 'main' panicked at aderyn_driver/src/process_auto.rs:96:17:
Error running solc command

The first step is to rename `remappings.txt` to `remappings` in the local copy of https://github.com/smartcontractkit/ccip/tree/ccip-develop/contracts

Then run the following command

`cargo run -- path/to/ccip/contracts --scope src/v0.8 --exclude "test/,tests/,lib/" -o ccip-play1.md`

### DIFF
--- a/aderyn_driver/src/process_auto.rs
+++ b/aderyn_driver/src/process_auto.rs
@@ -77,23 +77,33 @@ pub fn with_project_root_at(
     for (version, file_paths) in &compilation_groups {
         let solc = Solc::find_or_install_svm_version(version).unwrap();
         let solc_bin = solc.solc.to_str().unwrap();
-        dbg!(&solc_bin);
+        println!("{}", &solc_bin);
 
         let command = Command::new(solc_bin)
             .args(remappings.clone())
+            .arg("--base-path")
+            .arg(root_path)
+            .arg("--allow-paths")
             .arg("--ast-compact-json")
             .args(file_paths)
             .current_dir(root_path)
-            .stdout(Stdio::piped())
             .output();
 
         if let Ok(command) = command {
             if !command.status.success() {
+                let stdout = String::from_utf8(command.stdout.clone()).unwrap();
                 let msg = String::from_utf8(command.stderr).unwrap();
-                println!("stderr = {}, files={:?}", msg, file_paths);
+                println!("stderr = {}, stdout={}", msg, stdout);
                 println!("cwd = {}", root_path.display());
-                // dbg!(compilation_groups);
-                panic!("Error running solc command");
+                // dbg!(&compilation_groups);
+                let args = &file_paths
+                    .iter()
+                    .map(|x| x.to_string_lossy())
+                    .collect::<Vec<_>>()
+                    .join(" ");
+
+                println!("{} with version {}", args, version);
+                println!("\n\n\n");
             }
             let stdout = String::from_utf8(command.stdout).unwrap();
             let mut context = WorkspaceContext::default();
--- a/aderyn_driver/src/process_auto.rs
+++ b/aderyn_driver/src/process_auto.rs
@@ -49,10 +49,16 @@ pub fn with_project_root_at(
         if let Ok(version) = Solc::detect_version(source) {
             match compilation_groups.entry(format!("{}", version)) {
                 Entry::Occupied(mut o) => {
-                    o.get_mut().push(solidity_file.into());
+                    let s = solidity_file
+                        .strip_prefix(root_path.canonicalize().unwrap())
+                        .unwrap();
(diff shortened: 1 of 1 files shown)


TilakMaddy:
Okay, I might have found the issue here. (By doing some analysis after manually running the solc command)

Let me describe it.  The assumption of creating these compilation groups based on `Solc::detect_version()` (which looks at only the pragma) is incorrect!

It would have been correct if it also looked at the pragmas of the contracts being imported and then decided what's the latest version on disk it can use.

Example:
`F1.sol` has a pragma `^0.8.0`. It gets put into the `0.8.25` group.

But ...

If you `F1.sol` imports `F2.sol` which has a pragma `0.8.9`

Then you can say that the compilation with `0.8.25` doesn't work.

Hence it fails !!

TilakMaddy:
> Okay, I might have found the issue here. (By doing some analysis after manually running the solc command)
> 
> Let me describe it. The assumption of creating these compilation groups based on `Solc::detect_version()` (which looks at only the pragma) is incorrect!
> 
> It would have been correct if it also looked at the pragmas of the contracts being imported and then decided what's the latest version on disk it can use.
> 
> Example: `F1.sol` has a pragma `^0.8.0`. It gets put into the `0.8.25` group if `0.8.25` is available on disk
> 
> But ...
> 
> If you `F1.sol` imports `F2.sol` which has a pragma `0.8.9`
> 
> Then you can say that the compilation of `F1.sol` with `0.8.25` doesn't work.
> 
> Hence it fails !!

foundry_compilers natively tackles this issue by creating resolver graphs which scans for import statements to find out which version can be used to compile a solidity file. However, this can only be possible if the input to the solc binary is in standard json format because that will come with remappings packed into the json! (based on my understanding looking at the codebase) 

We, on the other hand supply the files as arguments to solc instead of the json input format. The advantage being we are  allowed to tack on the option `--ast-compact-json` (which isn't possible if the input is in standard json format)

I'm not sure yet what solution we can bring in here. But the problem seems to be clear

Ref to Std json format here https://docs.soliditylang.org/en/latest/using-the-compiler.html#compiler-input-and-output-json-description

TilakMaddy:
The reason why we didn't face this issue so far was because the contracts that were being imported had floating pragmas. So there were compatible with whatever version the parent file s pragma demanded

But in this repo, there is a sol file with fixed pragma which is being imported into a sol file with floating pragma.

That's the straw which breaks the camels back.

TilakMaddy:
ChainLinkUtil file with 0.8.9

aderyn_driver/src/process_auto.rs
```diff
@@ -71,22 +77,33 @@ pub fn with_project_root_at(
     for (version, file_paths) in &compilation_groups {
         let solc = Solc::find_or_install_svm_version(version).unwrap();
         let solc_bin = solc.solc.to_str().unwrap();
+        println!("{}", &solc_bin);
 
         let command = Command::new(solc_bin)
             .args(remappings.clone())
+            .arg("--base-path")
+            .arg(root_path)
```
alexroan, line 85:
```suggestion
            .arg(absolute_root_path_str)
```

I needed to change this for it to find the paths correctly.

alexroan (changes requested):

alexroan:
> sible if the input to the solc bin

It sounds like we have two options:
1. Copy the foundry method, switch to standard json input and benefit from the resolver graphs immediately.
2. Roll our own functionality to obtain the resolver graphs, and continue using file input.

Do you think we could try and benchmark both?

alexroan:
Also, is this related to #371 @TilakMaddy ?

TilakMaddy:
@alexroan We're not merging this PR!. It's just for us debugging : ) 

I have a couple of ideas that I can I'd like to try out this weekend : ) Stay tuned!

TilakMaddy:
Closing because no intention to merge this.
However we can still use this to debug : )

https://github.com/Cyfrin/aderyn/issues/393

TilakMaddy:
> Also, is this related to #371 @TilakMaddy ?

@alexroan The only thing I have done in #371 is check for bounds in the array before accessing the index. I didn't spend too much time thinking about the cause. But worth investigating - it's only that my current priority is fixing our method to calculate compilation groups : )

PR #398, opened 2024-05-06
Fix unused error false positives

alexroan:


alexroan:
cc: @malawadd, take a look at the false positives and the extra tests&code added here :)

PR #402, opened 2024-05-07
Icf/optimize context detection in parallel

TilakMaddy:
Initial State:

```python
for detector in detectors (parallel):
   for context in contexts: (sequential)
     `detector.detect(context)` 
```

With this PR

```python
for detector in detectors (parallel):
   for context in contexts: (parallel) <-----
     `detector.detect(context)` 
```


aderyn_core/src/detect/detector.rs
```diff
@@ -209,6 +209,12 @@ impl Display for IssueSeverity {
     }
 }
 
+impl dyn IssueDetector {
```
alexroan, line 212: Why did you need this @TilakMaddy ?

alexroan (approved):

PR #404, opened 2024-05-08
tx_origin_access_control

scab24:
New detector tasked with verifying if there are tx_origin_access_control statements in the code.


alexroan:
Hey @scab24 , it looks like the git history is slightly off. Could you base these changes from the current `dev` branch?

PR #407, opened 2024-05-08
Optimize layout order detector

TilakMaddy:


TilakMaddy:
Has forge changed it's behavior? Looks like we the open zeppelin libraries aren't accessible even when doing `forge build --ast`

alexroan (changes requested):
I think you're pulling in submodules from the internal-compilation-framework branch and it's screwing with forge here dude.

alexroan (approved):

PR #414, opened 2024-05-11
Icf/src

TilakMaddy:


.github/workflows/cargo.yml
```diff
@@ -150,7 +150,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: run
-          args: -- -o tests/ccip-functions-report-workflow.md tests/ccip-contracts/contracts --scope src/v0.8/functions/ --exclude "tests/,test/,testhelpers/,lib/,node_modules/,mocks/,vendor/" --skip-update-check
+          args: -- -o tests/ccip-functions-report-workflow.md tests/ccip-contracts/contracts --src src/v0.8/functions/ --exclude "tests/,test/,mocks/"  --skip-update-check
```
TilakMaddy, line 153: Notice how I am able to let go node_modules,lib, etc ...

TilakMaddy: We only have to exclude stuff that is inside `src`

TilakMaddy:
I'll try to send another PR attached to this as well. (For single file support) So merge that before this : )

aderyn/src/main.rs
```diff
@@ -34,6 +34,10 @@ pub struct CommandLineArgs {
     #[arg(short, long, default_value = "report.md")]
     output: String,
 
+    /// Path relative to project root, inside which solidity contracts will be analyzed
+    #[clap(long, use_value_delimiter = true)]
+    src: Option<Vec<String>>,
+
```
TilakMaddy, line 40: Hopefully this compelling enough !

TilakMaddy:
@alexroan Before merging this, get #424 in, to this branch (PR open)

alexroan (approved):

Issue #418, opened 2024-05-11
Requesting to follow this general rule of thumb when it comes to capturing `found_instances`


TilakMaddy:
This is w.r.t Auditor Mode (plus anything else in the future)

When we insert something `found_instances` in a struct, can we have it such that the reporter reporting the `found_instances` DOES NOT NEED the `context` to generate the report?

(So if we have to look-up the parent contract, or peek at the source code, etc...all that should be done and put into the `found_instances`. Not later)

This way we can run the reporter **once** (by giving the aggregation of found_instances as input) at the end to print everything. The alternative is - running the reporter once for each context (or) porting the context <-> found_instances matching logic to the reporter which frankly shouldn't be the reporters' job. Reporter should be focussed on how to "show" already baked stuff to the user. 

(Just saying)
No hard opinion

PR #419, opened 2024-05-11
Refactor/auditor mode (To be able to handle multiple contexts later)

TilakMaddy:
Fix https://github.com/Cyfrin/aderyn/issues/418


aderyn_core/src/lib.rs
```diff
@@ -52,12 +52,21 @@ where
...
+                return Some((detector.title(), detector.instances()));
+            }
+            None
+        })
+        .collect::<Vec<_>>();
+
+    for (detector_name, instances) in auditors_with_instances {
+        BasicAuditorPrinter::print(&instances, &detector_name);
+    }
+
```
TilakMaddy, line 69: Before: No potential to handle the printing of instances on multiple Contexts

Now: It is possible to run this on multiple contexys

TilakMaddy: (this will help me merge into ICF later if you approve this)

alexroan (approved):

aderyn_core/src/audit/auditor.rs
```diff
@@ -17,5 +27,31 @@ pub trait AuditorDetector: Send + Sync + 'static {
         String::from("Title")
     }
 
-    fn print(&self, _context: &WorkspaceContext) {}
+    fn instances(&self) -> Vec<AuditorInstance>;
+}
+
+pub trait AuditorPrinter {
```
alexroan, line 33: Hey @TilakMaddy , I think you'll need to rethink this for multiple workspaces. Each AuditorDetector will have its own print method, because every single one will have bespoke output and be formatted differently.

I'm reintroducing the print method in a separate PR after getting more input from the audit team. Stay tuned!

alexroan: #445

TilakMaddy: I mean so long as it is a table, we can make changes to the auditor instances to be any string array (Instead of specific fields that we have now)

TilakMaddy: Why don't we take it up after merging the current ICF to dev (or the equivalent - like adding process_foundry.rs, etc)?

alexroan: I know how to resolve this one now :) I'll throw up some changes for dev shortly

Issue #422, opened 2024-05-11
Foundry config can be loaded using foundry's repo


TilakMaddy:
We can load the config simply by using

https://github.com/foundry-rs/foundry/blob/503792a1dbadd901a4c02f6fcd1de1caff1573ff/crates/config/src/lib.rs#L135

Read how it works here --> https://github.com/foundry-rs/foundry/tree/master/crates/config

pros
-> If we use that, we don't have to parse `foundry.toml` ourselves and handle all those weird edge cases

cons
-> bundle size grows

Any thoughts ??

TilakMaddy:
I need to look into it more but it seems like we might just be able to support the current behavior of `dev` in `icf` if we take advantage of this `Config` struct in foundry's crates. (The challenge of requiring `remappings.txt` might be possible to offload becoz the methods in this `config` might just be able to give us the latent remappings PLUS the extra ones defined in the remappings sections for the relevant profile in foundry.toml)

TilakMaddy:
Will try to get a PR on to this https://github.com/Cyfrin/aderyn/pull/414 tomorrow by using defaults from `config` <3

TilakMaddy:
https://github.com/Cyfrin/aderyn/pull/424 💥

TilakMaddy:
Closing because merged !

PR #424, opened 2024-05-12
Use foundry config crate to detect src, remapppings, etc (To offload parsing of foundry.toml to forge maintainers)

TilakMaddy:
@alexroan With this PR we will be mostly compatible with `dev` in terms of dev ex
(The exception is that we must have `remappings.txt` in the project root - which I will try to get rid of as well)

UPDATE:
PR #425 takes care of removing the requirement for `remappings` file for a foundry project

Try running
```
FOUNDRY_PROFILE=uniswap cargo run tests/contract-playground/ -o tests/uniswap_profile.md
```
It will auto detect src, scope and exclude  (with the ability to override by cli args)

So there's a lot of cases (if we do it ourselves)
* You have to look at foundry.toml
* You have to look at env variables
* You have to consider only the relevant profile

So it's better we just rely on the config crate (I think)

---------

Prereq (to merge that first)
https://github.com/Cyfrin/aderyn/pull/425


TilakMaddy:
Note
In cargo.yml, we now install the nightly toolchain instead of the stable one (only for lints section) 

This is because we need to be able to run `cargo fmt --all -- --check` with the ignore foundry option and that only works with the nightly toolchain. 
(Btw ignore is fed from the `rustfmt.toml` file :P )

TilakMaddy:
Fix https://github.com/Cyfrin/aderyn/issues/422

alexroan (approved):

PR #434, opened 2024-05-13
Debouncer - refactor code (small wait time)

TilakMaddy:
Pls ignore the branch name - this is not actually dynamically adjusted. (I tried but didn't see the point in doing so. Hence, I'm just refactoring)

New wait time is 50ms (in between changes - small or big)

### DIFF
--- a/aderyn/src/main.rs
+++ b/aderyn/src/main.rs
@@ -5,7 +5,7 @@ use serde::Deserialize;
 use serde_json::Value;
 use std::{
     path::{Path, PathBuf},
-    time::{self, Duration},
+    time::Duration,
 };
 use strum::IntoEnumIterator;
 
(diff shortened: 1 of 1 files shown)


aderyn/src/main.rs
```diff
@@ -285,29 +269,14 @@ fn main() {
...
-                }
-                println!();
-            }
+
+            debounce_and_run(
+                || {
+                    driver::drive(args.clone());
+                },
+                &args,
+                Duration::from_millis(50),
```
TilakMaddy, line 278: Here

aderyn/src/main.rs
```diff
@@ -233,43 +233,27 @@ fn main() {
...
                             }
-                            Err(errors) => errors.iter().for_each(|error| println!("{error:?}")),
-                        }
-                        println!();
-                    }
+
+                            driver::drive_with(new_args.clone(), subscriptions);
+                        },
+                        &new_args,
+                        Duration::from_millis(50),
```
TilakMaddy, line 255: And here

alexroan (approved):
Nice! the watch time is much better here. @TilakMaddy 

After this, it looks like the slowest part is now:
```
Resolving sources versions by graph ...
```

PR #439, opened 2024-05-13
Merge dev to icf again

TilakMaddy:
This PR Resolve merge conflicts from latest dev into ICF

PS:
I kind of butchered sarif-reporter :P (to make it handle multiple contexts)

@alexroan Please look into sarif report here. It looks like there are more instances captured after I made the adjustment for multiple contexts. (I even used `dedup()` on locations)

(OR) verify that the sarif-report.md is correct !

~UPDATE:
Problem found - dedup doesn't work as expected. Will have to give a custom key for deduplication and try it. : )~


aderyn_core/src/report/sarif_printer.rs
```diff
@@ -0,0 +1,213 @@
...
+                    logical_locations: None,
+                    relationships: None,
+                    message: None,
+                };
+                locations.push(location);
+            }
+        }
+    }
+    locations.dedup();
+    locations
```
TilakMaddy, line 212: I'm dedupping here

TilakMaddy:
Wait NOW I think i got it - we're facing a similar issue described here - https://github.com/Cyfrin/aderyn/issues/418

In the reporter logic, we can't rely on the `context` to give us things like `offset_len`, etc !!

alexroan (approved):

PR #445, opened 2024-05-13
Feature: Auditor-detector: public/external functions without `msg.sender` checks

alexroan:


aderyn_core/src/lib.rs
```diff
@@ -52,21 +52,24 @@ where
 }
 
 fn run_auditor_mode(context: &WorkspaceContext) -> Result<(), Box<dyn Error>> {
-    let auditors_with_instances = get_auditor_detectors()
```
alexroan, line 55: @TilakMaddy hopefully this makes it easier to build more auditor-mode detectors with ICF

TilakMaddy: @alexroan I'm yet to look into it ... but on a side note I was wondering if we can bring sarif and auditor mode support to ICF *after* the merge. So basically if you have `--icf -o report.sarif` we will say sarif mode is not supported yet with `--icf`. ...and likewise if you have `--icf --auditor-mode` we will say `auditor-mode` is not supported yet.

reports/report.md
```diff
@@ -957,25 +999,7 @@ The `ecrecover` function is susceptible to signature malleability. This means th
 
 
 
-## L-4: Deprecated OpenZeppelin functions should not be used
```
alexroan, line 960: Why is this being deleted??? 🤔

reports/report.json
```diff
@@ -843,7 +862,7 @@
   },
   "issue_count": {
     "high": 6,
-    "low": 23
+    "low": 22
```
alexroan, line 865: Why??

reports/report.md
alexroan: OMG I've found the reason!

The ID `5463` (which is the ID of the source_unit for DeprecatedOZFunctions.sol) is duplicated! 0.8.19, 0.8.20 AND 0.8.25 all product the same ID for different ASTNodes, so they're being overwritten in the WorkspaceContext.

Holy crap @TilakMaddy , we definitely need the multiple WorkspaceContexts!

alexroan:
On hold until ICF is ready.

reports/report.md
TilakMaddy: @alexroan Is this the only thing blocking?

TilakMaddy:
Which one is the latest? `AttackSurface` or `CallsAndDelegates` ?

reports/report.md
TilakMaddy: @alexroan https://github.com/Cyfrin/aderyn/pull/459  <---- ICF doesn't do this. Look here !!

aderyn_core/src/audit/public_functions_no_sender_checks.rs
```diff
@@ -0,0 +1,134 @@
...
+    },
+};
+use std::{cmp::Ordering, collections::BTreeSet, error::Error};
+
+#[derive(Clone, Eq, PartialEq)]
+pub struct NoChecksInstance {
+    pub contract_name: String,
+    pub function_name: String,
+    pub function_kind: FunctionKind,
+}
```
TilakMaddy, line 21: Please change this to String

TilakMaddy:
@alexroan  you can continue your work here https://github.com/Cyfrin/aderyn/pull/459

PR #452, opened 2024-05-18
Icf/fix sarif results

TilakMaddy:


aderyn_core/src/report/sarif_printer.rs
```diff
@@ -28,7 +28,7 @@ impl ReportPrinter<()> for SarifPrinter {
         &self,
         writer: W,
         report: &Report,
-        contexts: &[WorkspaceContext],
+        _: &[WorkspaceContext],
         _: PathBuf,
```
TilakMaddy, line 32: I have underscored this parameter because I want you to not make use of `context` while printing. 

Later I will hard remove it :) It is not good practice and causes problems when handling multiple contexts.

cc: @alexroan

PR #455, opened 2024-05-18
CI Legacy reports generation workflow (when there is no `--icf` flag)

TilakMaddy:
There is a new job in CI called `Check Reports (Legacy)` 

That's where all the current dev tests should go.

cc: @alexroan


TilakMaddy:
CI tests from dev imported ✅

PR #459, opened 2024-05-18
ICF Feature: Auditor-detector: public/external functions without msg.sender checks

TilakMaddy:
[https://app.warp.dev/block/otsvu5ksb7reSyvNLm58W6](https://app.warp.dev/block/otsvu5ksb7reSyvNLm58W6)

Checkout this branch
Run `makes setup`

Then chmod +x for all scripts in `cli/`

Then run 
```
cargo run -- tests/contract-playground/ --auditor-mode --icf
```


TilakMaddy:
#445

TilakMaddy:
Looks like we have to add a filter by matching contract name option...But let's do it after the merge Alex : )

TilakMaddy:
Let me take care of merging this after you are done : )

TilakMaddy:
We expect the job Reports vs Legacy Reports to fail here !

TilakMaddy:
Alex see if you can get the `Check Reports (legacy)` job to pass! 

We do expect the `Reports vs Legacy Reports`  to fail. So you can leave that alone : )

TilakMaddy:
Apparently on adding the new solidity file, 

we lose the followin
```
## L-4: Deprecated OpenZeppelin functions should not be used

Openzeppelin has deprecated several functions and replaced with newer versions. Please consult https://docs.openzeppelin.com/

- Found in src/DeprecatedOZFunctions.sol [Line: 22](../tests/contract-playground/src/DeprecatedOZFunctions.sol#L22)

	```solidity
	        _setupRole(role, account);
	```

- Found in src/DeprecatedOZFunctions.sol [Line: 27](../tests/contract-playground/src/DeprecatedOZFunctions.sol#L27)

	```solidity
	        token.safeApprove(spender, value);
	```
```

alexroan:
> Apparently on adding the new solidity file,
> 
> we lose the `Deprecated OpenZeppelin functions should not be used` when run locally .... and we lose the `ERC721::_mint()` when CI runs it
> 
> That's why we are seeing 22 lows in both cases + so they fail **check reports (legacy)** job
> 
> This is slightly frustrating because even if you manage to generate the same thing locally and on CI, I may not !

Yep, this is due to Node IDs overlapping for different solc versions, in the same WorkspaceContext.

The fact that it only happens in legacy mode is a good thing 👍

alexroan:
Closing in favour of #479

Issue #492, opened 2024-06-01
--highs-only


alexroan:
**Is your feature request related to a problem? Please describe.**
Provide a highs-only flag

PR #497, opened 2024-06-02
Added `--highs-only` 

TilakMaddy:
Fix https://github.com/Cyfrin/aderyn/issues/492


aderyn_driver/src/driver.rs
```diff
@@ -33,6 +33,13 @@ pub fn drive(args: Args) {
     drive_with(args, get_all_issue_detectors());
 }
 
+pub fn drive_highs_only(args: Args) {
```
alexroan, line 36: Does it make sense to introduce another drive function? It feels like this path should be covered in `drive_with`. What do you think?

TilakMaddy: Yes. I think so too. Let me do that

TilakMaddy: Done ✅ - Removed the extra function by including in `highs_only` as a field in `Args` 

Also included test to generate json report for highs-only

PR #499, opened 2024-06-03
Graceful failure with good error message when no solidity files found !

### LINKED ISSUE #543: Tests for graceful exit condition on empty contexts .
https://github.com/Cyfrin/aderyn/pull/499

TilakMaddy:
fix #543


aderyn_driver/src/foundry_compiler_helpers.rs
```diff
@@ -17,6 +18,10 @@ pub fn get_compiler_input(root: &Path) -> CompilerInput {
         .into_iter()
         .filter(|c| c.language == *"Solidity")
         .collect::<Vec<_>>();
+    if solidity_files.is_empty() {
+        eprintln!("No solidity files found in {}!", root.to_string_lossy());
+        exit(1);
+    }
     let solidity_files = &solidity_files[0]; // No Yul Support as of now
```
TilakMaddy, line 25: This line panics if the above check isn't made

aderyn_driver/src/driver.rs
```diff
@@ -115,6 +115,11 @@ fn make_context(args: &Args) -> WorkspaceContextWrapper {
     let mut contexts: Vec<WorkspaceContext> =
         process_auto::with_project_root_at(&root_path, &src, &exclude, &remappings, &include);
 
+    if contexts.iter().all(|c| c.src_filepaths.is_empty()) {
+        eprintln!("No solidity files found in given scope!");
```
alexroan, line 119: Can you introduce a test for this please @TilakMaddy ?

aderyn_driver/src/foundry_compiler_helpers.rs
```diff
@@ -17,6 +18,10 @@ pub fn get_compiler_input(root: &Path) -> CompilerInput {
         .into_iter()
         .filter(|c| c.language == *"Solidity")
         .collect::<Vec<_>>();
+    if solidity_files.is_empty() {
```
alexroan, line 21: Test here too please

alexroan (changes requested):

alexroan:
@TilakMaddy can you please add those tests in here? :)

TilakMaddy:
Sorry, I forgot about this. Will add tests shortly : )

aderyn_driver/src/foundry_compiler_helpers.rs
TilakMaddy: This is to check for safe array index access in the next line ` let solidity_files = &solidity_files[0]; `

We must ensure vector is not empty if we want to access index 0 of the vector : )

aderyn_driver/src/driver.rs
TilakMaddy: What might we test for in this case ?

Say we form a contexts vector from a directory of solidity files but with no solidity files matching the given scope.

Like, `let mut contexts: Vec<WorkspaceContext> =
        process_auto::with_project_root_at(&root_path, &src, &exclude, &remappings, &include);`

Then test for `contexts.iter().all(|c| c.src_filepaths.is_empty()) == true`  is alright?

TilakMaddy: DONE ✅

aderyn_driver/src/foundry_compiler_helpers.rs
TilakMaddy: Idea - 💡 Let me write a test for this by having a directory with no solidity files at all whatsoever.

alexroan (approved):

Issue #500, opened 2024-06-03
AST Error - SourceUnit lacks top level `FunctionDefinition` hence `ErrorDefinition` falsely deserializes objects


TilakMaddy:
Functions defined freely outside of contracts in a helper file are read as `ErrorDefinition` instead  `FunctionDefinition` 

Problems:

1) It's a semantic error (this is bad enough)
2) The function's body doesn't make it into the AST (because ErrorDefinitions don't have bodies)

BTW, it deserializing fails silently like this. You won't see any panics

alexroan:
Can you show an example of this @TilakMaddy ?

TilakMaddy:
@alexroan 

You can try this

```
git clone https://github.com/transmissions11/solmate
forge build
aderyn
```

TilakMaddy:
@alexroan  

UPDATE:

I think I just fixed it :) 😉 
Will send a PR

PR #508, opened 2024-06-03
Fix/ Added support for Free Functions to AST

TilakMaddy:
Fix https://github.com/Cyfrin/aderyn/issues/500


aderyn_core/src/ast/source_units.rs
```diff
@@ -8,6 +8,7 @@ use std::{collections::HashMap, io};
 #[derive(Clone, Debug, Deserialize, Eq, Serialize, PartialEq, Hash)]
 #[serde(untagged)]
 pub enum SourceUnitNode {
+    FunctionDefinition(FunctionDefinition),
```
TilakMaddy, line 11: This is strategically placed at the top (or I should say above ErrorDefinition to be specific) because its' fields are a superset of ErrorDefinition and therefore if Rust can deserialize the object into this it will do so. Otherwise, ErrorDefinition would be the first satisfied case and would stop the program from detecting free functions

TilakMaddy:
Whilst this one is fixed, I think I'll take a closer look at these AST Nodes a little bit more. I expect to find more such bugs 🐞 
We need to at least catch up to 0.8.26 solidity : )

alexroan (approved):
Great stuff @TilakMaddy

PR #509, opened 2024-06-03
Fix Detector/ Do not capture unused public functions if they live in an abstract contract

TilakMaddy:
.


alexroan (commented):
Approving this.

However, In future, let's remember to write specific unit tests for small changes in functionality like this.

The sablier.md report acts as an integration test atm, which upon reviewing, a developer might miss important changes to it with large diffs. 
A unit test, on the other hand, would identify exactly what changed and how to fix it, making future development practice more robust @TilakMaddy .

aderyn_core/src/detect/low/useless_public_function.rs
```diff
@@ -91,4 +103,19 @@ mod useless_public_function_tests {
         // assert that the detector returns the correct description
         assert_eq!(detector.description(), String::from("Instead of marking a function as `public`, consider marking it as `external` if it is not used internally."));
     }
+
+    #[test]
+    #[serial]
+    fn test_useless_public_functions_does_not_capture_abstract_contract_functions() {
```
alexroan, line 109: Unit test added here @TilakMaddy

TilakMaddy:
> Approving this.
> 
> However, In future, let's remember to write specific unit tests for small changes in functionality like this.
> 
> The sablier.md report acts as an integration test atm, which upon reviewing, a developer might miss important changes to it with large diffs. A unit test, on the other hand, would identify exactly what changed and how to fix it, making future development practice more robust @TilakMaddy .
> 
> I'm adding this test in a final commit here before merging :)

Understood! Thanks Alex

PR #513, opened 2024-06-04
Adding prb math to CI 

TilakMaddy:
PRB Math uses a lot of the new features from newer solidity versions.

* Used as health-check for AST Apis


TilakMaddy:
I have managed to make it work on PRB Math. Mainly the change was with `UsingForDirective`

TilakMaddy:
NOTE - We still have a long way to go with AST :P It's not looking good and I also don't want to make this PR too long.

alexroan (approved):

Issue #557, opened 2024-06-23
Something doesn't feel right with the speed of execution of aderyn release binaries


TilakMaddy:
Let me explain 

Choose a large repo like lssvm2.

**First try**

Run `cargo build --verbose --locked --release --package aderyn --target x86_64-apple-darwin` locally
Then run `./target/x86_64-apple-darwin/release/aderyn path/to/lssvm2 -o option1.md`

(Measure the time)

**Second try**

Do the whole `cyfrinup` thing to get the Aderyn binary (or manually download it and unzip it from releases)
Then run `aderyn path/to/lssvm2 -o option2.md`

(Measure the time)

### I find that the First try takes **significantly** less time in resolving sources by graph step and is almost instantaneous whereas the second try seems to behave more like the dev version where it is taking a few seconds in the graph resolution step

@alexroan Please 🙏 can you try this experiment too ?

alexroan:
I ran this:

`cargo build --verbose --locked --release --package aderyn`
`./target/release/aderyn path/to/lssvm2`

and

`cargo build --locked --release --package aderyn --target aarch64-apple-darwin
`./target/aarch64-apple-darwin/release/aderyn /path/to/lssvm2/`

And this was very similar speed to `cyfrinup && aderyn`.

Could you dig into how your release binary is being built differently to the x86_64 darwin build in CI please @TilakMaddy ? We might be missing something.

alexroan:
Fixed, will be released in v0.1.4

TilakMaddy:
Finally !!! Now I'm able to run the latest fastest aderyn with cyfrinup on my mac.  Can't wait for nightly builds <3

PR #562, opened 2024-06-24
MacOS: x86_64 binary - use macos with intel chip as runner

TilakMaddy:
Reference - 
https://github.com/actions/runner-images?tab=readme-ov-file

Currently, we are building x86_64 binaries in an arm64 based mac runner. Plausible cause for #557 

So this PR changes the runner to an intel based mac based on the reference above

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,7 +24,7 @@ jobs:
             runner: windows-latest
             target: x86_64-pc-windows-msvc
           - name: macos-amd64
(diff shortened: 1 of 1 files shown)


alexroan:
Based on [these docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#about-macos-larger-runners), should we run the arm build on the `xlarge` runner @TilakMaddy ?

alexroan (approved):

PR #577, opened 2024-07-10
Fix TempleDao repo issues

### LINKED ISSUE #574: Aderyn crashes for TempleDao
Repo source:
https://github.com/TempleDAO/temple/tree/templegold/protocol/contracts

Command:
```
cargo run path/to/temple/protocol/ -o temple.md
```

Output:
```
   Blocking waiting for file lock on build directory
   Compiling aderyn_core v0.1.5 (/Users/tilakmadichetti/Documents/OpenSource/realaderyn/aderyn_core)
   Compiling aderyn_driver v0.1.5 (/Users/tilakmadichetti/Documents/OpenSource/realaderyn/aderyn_driver)
   Compiling aderyn v0.1.5 (/Users/tilakmadichetti/Documents/OpenSource/realaderyn/aderyn)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 16.42s
     Running `target/debug/aderyn ../temple/protocol/ -o temple.md`
Root: "/Users/tilakmadichetti/Documents/OpenSource/temple/protocol", Src: Some(["contracts"]), Include: None, Exclude: Some(["test/forge/", "script/", "node_modules/", "lib/"])
Resolving sources versions by graph ...
Compiling 13 files with Solc 0.5.16+commit.9c3226ce.Darwin.appleclang
Compiling 10 files with Solc 0.6.6+commit.6c089d02.Darwin.appleclang
Compiling 174 files with Solc 0.8.20+commit.a1b79de6.Darwin.appleclang
Compiling 154 files with Solc 0.8.26+commit.8a97fa7a.Darwin.appleclang
Get Detectors
Running 30 detectors
Detectors run, processing found issues
Found issues processed. Printing report
thread 'main' panicked at aderyn_core/src/report/mod.rs:106:70:
called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }
```

Helpful analysis (bug cause):
```
[aderyn_core/src/report/mod.rs:107:25] contract_path = "contracts/amm/TempleStableAMMRouter.sol"
[aderyn_core/src/report/mod.rs:107:25] line_no = 148
[aderyn_core/src/report/mod.rs:107:25] src_location = "-1:-1"
```

alexroan:
Fixes #574 

* Unreachable code in `variables.rs` defaults to `Mutability:immutable`
* `get_node_sort_key` uses `name_location` only if it doesn't contain `-1`. Otherwise it uses `src`
* Added Templegold to CI


alexroan:
After fixing the unreachable code, I now get the same error as described in #574 .

The error occurs in the `inconsistent-type-names` detector. It's capturing something that has a source location of "-1:-1:-1".

TilakMaddy:
Alex, this is not a correct fix for state mutability. Please dont push it. Here's why -

For older versions of solidity, state mutability can be none while constant and mutable properties carry the information of mutability. Therefore the helper function is required to abstract that away.

One potential fix is to return None instead of unreachable.

TilakMaddy:
State mutability can be none when it doesn't apply .for example in interface function definitions

alexroan:
Aligned, I'll use None 👍

alexroan:
@TilakMaddy can you please review this? :)

TilakMaddy:
Question - 
1) How do you know that it's immutable and not just None  ? Solidity immutable keyword was added in 0.6 . Therefore this cannot be immutable imo . It has to be none

 (We must change the return type of the helper function to be option<state mutability>)

2) what about name location then .. Looks like it's out of the picture now. (If thats  true, why don't we keep the transformations, except that when "-1" is found in name location, you return source location)

TilakMaddy:
Question - 
1) How do you know that it's immutable and not just None  ? Solidity immutable keyword was added in 0.6 . Therefore this cannot be immutable imo . It has to be none

 (We must change the return type of the helper function to be option<state mutability>)

2) what about name location then .. Looks like it's out of the picture now. (If thats  true, why don't we keep the transformations, except that when "-1" is found in name location, you return source location)

alexroan:
> How do you know that it's immutable and not just None ? Solidity immutable keyword was added in 0.6 . Therefore this cannot be immutable imo . It has to be none

Good point. Changed to return Optional, and return None in the `else` case.

> what about name location then .. Looks like it's out of the picture now. (If thats true it will be a problem, why don't we keep the transformations, except that when "-1" is found in name location, you return source location. Otherwise, just return name location)... It's important for vscode extension to not highlight the whole contract for example, instead just highlight the name

Let's think about this as a follow-up. I'd be interested to see how it affects report output.

PR #582, opened 2024-07-12
Detector: Incorrect Assembly Shift Parameter Order

alexroan:


aderyn_core/src/detect/high/incorrect_shift_order.rs
```diff
@@ -0,0 +1,90 @@
...
+impl IssueDetector for IncorrectShiftOrderDetector {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
+        let yul_function_calls = context.yul_function_calls();
+        for yul_function_call in yul_function_calls {
+            if yul_function_call.function_name.name == "shl"
+                || yul_function_call.function_name.name == "shr"
+            {
+                if let YulExpression::YulLiteral(_) = &yul_function_call.arguments[1] {
+                    capture!(self, context, yul_function_call);
+                }
```
TilakMaddy, line 30: Yul function calls don't have IDs, how are we able to capture these ??  (Confused) 😕

TilakMaddy: (I'm trying to adjust the big refactor PR for Yul stuff)

TilakMaddy: Ahh I got it - the capture! macro was changed to insert `0` as the id if it's none !

PR #591, opened 2024-07-16
New Detector: (Low) Uninitialized state vars consumed

TilakMaddy:
Pre-req : #594


TilakMaddy:
(Unable to replicate this error locally)

aderyn_core/src/detect/low/uninitialized_state_variable.rs
```diff
@@ -0,0 +1,150 @@
...
+#[derive(Default)]
+pub struct UninitializedStateVariableDetector {
+    // Keys are: [0] source file name, [1] line number, [2] character location of node.
+    // Do not add items manually, use `capture!` to add nodes to this BTreeMap.
+    found_instances: BTreeMap<(String, usize, String), NodeID>,
+}
+
+impl IssueDetector for UninitializedStateVariableDetector {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
+        /*
```
alexroan, line 23: I like this plan! It sounds like this will correctly _not_ capture state variables that are declared, but initialized in a contract that extends the contract the variable was declared. For example:

```sol

contract base{
    uint256 public myVar;
}

contract extension is base {
    constructor() {
        myVar = 4;
    }
}
```

alexroan (approved):

aderyn_core/src/detect/low/uninitialized_state_variable.rs
alexroan: In fact, could you add a test for this please @TilakMaddy ?

alexroan (changes requested):

aderyn_core/src/detect/low/uninitialized_state_variable.rs
```diff
@@ -0,0 +1,150 @@
...
+                .nodes
+                .get(&id)
+                .inspect(|&x| capture!(self, context, x));
+        }
+
+        Ok(!self.found_instances.is_empty())
+    }
+
+    fn severity(&self) -> IssueSeverity {
+        IssueSeverity::Low
```
alexroan, line 83: Why low instead of high?

aderyn_core/src/detect/low/uninitialized_state_variable.rs
TilakMaddy: Test worked! It has correctly not captured it :)

aderyn_core/src/detect/low/uninitialized_state_variable.rs
TilakMaddy: Yes, it's actually high. I'll change it

TilakMaddy:
What is up with these conflicts lol? `git` CLI literally shows that everything is resolved when  `git merge dev` is run locally but here, Github says there are conflicting files

aderyn_core/src/detect/low/uninitialized_state_variable.rs
TilakMaddy: ✅

alexroan (approved):

reports/report.md
```diff
@@ -1143,7 +1145,96 @@ If the length of a dynamic array (storage variable) directly assigned to, it may
...
+- Found in src/eth2/DepositContract.sol [Line: 63](../tests/contract-playground/src/eth2/DepositContract.sol#L63)
+
+	```solidity
+	    bytes32[DEPOSIT_CONTRACT_TREE_DEPTH] branch;
+	```
+
+- Found in src/eth2/DepositContract.sol [Line: 66](../tests/contract-playground/src/eth2/DepositContract.sol#L66)
+
+	```solidity
+	    bytes32[DEPOSIT_CONTRACT_TREE_DEPTH] zero_hashes;
```
alexroan, line 1230: This seems to catch an array that actually is assigned in the constructor. Is this an edge case not accounted for?

alexroan (changes requested):

reports/report.md
TilakMaddy: Good catch! This is a fixed size array. It also needs to be accounted for

.github/workflows/cargo.yml
```diff
@@ -268,6 +255,50 @@ jobs:
...
+
+  third-party-reports:
+    name: Check 3rd party Reports
+    runs-on: ubuntu-latest
+    steps:
+      - name: foundry-toolchain
+        uses: foundry-rs/foundry-toolchain@v1.2.0
+
+      - name: Checkout sources
+        uses: actions/checkout@v2
```
alexroan, line 267: @TilakMaddy 

Can you:
- Merge dev into here
- Remove the change to this file

See if it passes?

alexroan (changes requested):

alexroan (approved):

PR #592, opened 2024-07-16
Detector: State variable shadowing

alexroan:
Thoughts: I hate it.

Lots of recursion through the inheritance tree, and checking the solidity compiler version going on.

@TilakMaddy if you can think of a better algorithm to find this issue, please let me know and I will amend it!


TilakMaddy:
I like the 2 points that you have mentioned for optimization. Can I move this / clone one up in the auditor mode and take it up there ? (I'll have the flexibility to create preprocessing trees and traverse without worrying about performance too much in the get go)

(I'm not too clear about the override keyword for variables I thought that was only for functions :P )

TilakMaddy:
UPDATE:
🤔 just thinking about an estimate of depth of inheritance in contracts - maybe like 3 or 4 levels deep max ? If so, then is too bad we are repeating recursion ?

PR #594, opened 2024-07-17
Fix testing infra bugs

TilakMaddy:
Before: Support for older versions solidity had a bug 
Now: Imported the solution from `process_auto` which has proved to work so far (so it should unblock detectors tests)

Changed made to - 

`load_solidity_source_unit`
`load_multiple_solidity_source_units_into_single_context`


TilakMaddy:
CI Run on commit d3839e9 failed due to ExcutableFileBusyError (Which, when happened during ICF was fixed by aadding duplicate tests runs before the actual cargo test)

NOTE - Unable to replicate that error locally

alexroan:
This was merged in a separate PR!

PR #597, opened 2024-07-17
Functions that send native eth without having msg.sender checks

TilakMaddy:
This creates a callgraph. Because we are unsure of performance, etc. ... I have added this to `--auditor-mode` 

`cargo run tests/contract-playground --auditor-mode`

Let's play with this for a while before converting it to a detector 

UPDATE:

Now it's converted to deetctor


aderyn_core/src/audit/send_ether_no_checks.rs
TilakMaddy, line 1: New auditor detector here :)

aderyn_core/src/audit/send_ether_no_checks.rs
```diff
@@ -0,0 +1,132 @@
...
+#[derive(Default)]
+pub struct SendEtherWithoutMsgSenderChecksDetector {
+    found_instances: Vec<SendEtherNoChecksInstance>,
+}
+
+impl AuditorDetector for SendEtherWithoutMsgSenderChecksDetector {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
+        for func in helpers::get_implemented_external_and_public_functions(context) {
+            let mut tracker = MsgSenderAndCallWithValueTracker::default();
+            let investigator = SimpleInvestigator::for_node(func, context)?;
```
alexroan, line 37: For my context: Does this create the CFG only on request from the Detector?

Is there a way we can do it (efficiently) globally as a preprocessing step so that all detectors can access it by default? How would that affect performance?

alexroan (changes requested):
We want this to be a normal detector, so let's move it there.

Completely agree with concerns about performance. My thinking is that we'll do a sprint later, once we have more detector tickets completed, which will be focussed on performance, efficiency, etc.

aderyn_core/src/audit/investigators.rs
```diff
@@ -0,0 +1,208 @@
...
+impl std::error::Error for Error {}
+
+/// It creates simple call graph from a given [`FunctionDefinition`] and provides a hook to the
+/// visitor for investigating in-the-path nodes that may be of interest.
+///
+/// SimpleInvestigator does a Depth First Search to recursively visit the following types of Nodes
+/// - [`ModifierDefinition`]
+/// - [`FunctionDefinition`]
+///
+/// NOTE: This is NOT a Control-Flow-Graph so that means it does not promise *any* kind of order
```
TilakMaddy, line 53: @alexroan Unfortunately this is not CFG :P It's only a callgraph.

aderyn_core/src/audit/send_ether_no_checks.rs
TilakMaddy: 💯 Global preprocessing is the goal but I thought Auditor-mode is a way to introduce it to aderyn, so that we can play with it for a while and see what can be done :)

TilakMaddy: Yes, it creates the Callgraph on request from the detector

TilakMaddy: ✅ It's now a detector

reports/report.md
```diff
@@ -1659,6 +1678,24 @@ Instead of marking a function as `public`, consider marking it as `external` if
 	    function onlyOwnerManualIfRevertCheck() public view {
 	```
 
+- Found in src/auditor_mode/send_ether_no_checks/Tower.sol [Line: 19](../tests/contract-playground/src/auditor_mode/send_ether_no_checks/Tower.sol#L19)
```
alexroan, line 1681: Can we make these external so that they aren't caught here?

reports/report.md
```diff
@@ -1222,6 +1224,23 @@ Description of the high issue.
 
 
 
+## H-19: Sending native Eth is not protected from these functions.
+
+Description of the high issue.
+
+<details><summary>1 Found Instances</summary>
+
+
```
alexroan, line 1233: Only one instance found? Should we expect more?

aderyn_core/src/audit/investigators.rs
```diff
@@ -0,0 +1,208 @@
+//! This module helps with strategies on performing different types of investigations.
```
alexroan, line 1: Do you think we should move this to a different folder? Maybe `browser` or `context` or its own new one?

reports/report.md
TilakMaddy: No, just one I think.

aderyn_core/src/audit/investigators.rs
TilakMaddy: `browser` would suitable.

TilakMaddy:
This detector will be part of #605

PR #600, opened 2024-07-19
Controlled delegatecall destination check !

TilakMaddy:
Command to run:
`cargo run -- tests/contract-playground/ --auditor-mode`

(If I can get of a review the logic + output for now, it would help. Will convert to detector later.  I just want thoughts on the findings in the table and the limitations described in the Solidity file tagged below are OK)


tests/contract-playground/src/auditor_mode/delegate_call_no_checks/DelegateCallWithoutAddressChecks.sol
```diff
@@ -0,0 +1,52 @@
...
+    // either in the constructor or in other methods with modifiers.
+    address public manager;
+
+    constructor() {
+    
+    }
+
+    function delegate1(address to, bytes memory data) external {
+        to.delegatecall(data); // `to` is not protected, therefore BAD
+    }
```
TilakMaddy, line 17: Here. we can see that this function doesn't perform any checks yet, makes a delegate call on an address that is not a state variable.

So it's considered BAD (and we capture it)

tests/contract-playground/src/auditor_mode/delegate_call_no_checks/DelegateCallWithoutAddressChecks.sol
```diff
@@ -0,0 +1,52 @@
...
+                isLegit = true;
+            }
+        }
+        require(isLegit);
+        _;
+    }
+
+    function delegate3(address to, bytes memory data) external isAllowed(to) {
+        to.delegatecall(data); // `to` is protected, therefore GOOD
+    }
```
TilakMaddy, line 37: This function performs a delegate call on a non state variable. + performs checks on address (look inside the `isAllowed` modifier) 

So it's good. We don't capture it

alexroan:
Let's make this a normal detector

TilakMaddy:
This detector will be part of #605

PR #602, opened 2024-07-19
Extend support for `reference_declarations()` on `Expression`

TilakMaddy:
I find that while writing detectors, we often tend to do 

`if let Expression::Identifier(id) = expr { ..  ...id.referenced_declaration .. } `

`if let Expression::MemberAccess(m) = expr { ..  ...m.referenced_declaration .. } `

Which maybe correct in many places, but a lot of times we just want to know what the expression is referencing, so it's easier and more scalable to have a `referenced_declarations()` on the Expression itself. We've had one so far but it was far out of date. This PR extends support to it

### DIFF
--- a/aderyn_core/src/ast/impls/own/expressions.rs
+++ b/aderyn_core/src/ast/impls/own/expressions.rs
@@ -102,16 +102,14 @@ impl Expression {
                     function_call_ops
                         .options
                         .iter()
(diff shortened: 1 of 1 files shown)


TilakMaddy:
Closing this in favor of `ExtractReferencedDeclaration::from(...)` which would be way more scalable and easier to maintain

PR #604, opened 2024-07-19
Performance boost for aderyn 

TilakMaddy:
* Alternative allocator kicks in when available based on platform https://crates.io/crates/tikv-jemallocator (For all modes)
* Codegen units = 1 (For release mode)
* Link time optimization = "fat" (For release mode)

TODO:
change to mimalloc instead of jemalloc

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ exclude = [
     "foundry",
 ]
 
+
 [profile.release]
 codegen-units = 1
 lto = true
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -11,4 +11,9 @@ resolver="1"
...
     "bot_ci_cd",
     "foundry",
-]
\ No newline at end of file
+]
+
+
+[profile.release]
+codegen-units = 1
+lto = true
```
TilakMaddy, line 19: I think we'll keep this only for now. Regarding the heap allocators I need to experiment more

PR #605, opened 2024-07-20
Global pre-processing for leveraging callgraphs 

TilakMaddy:
CallGraphs can now be leveraged with help of `StandardInvestigator` - I have tried to make as many default assumptions as possible to keep the interface simple to use.

Job of the investigator is to derive a subgraph from the master workspace callgraph and provide hooks to interact with functions and modifiers while it does the following:
* Navigate upstream 
* Navigate downstream 
* Navigate upstream's downstream-side-effects (These are nodes that are downstream from the upstream nodes but are not upstream themselves)

_PS: You can provide a navigation style to save time and be efficient in terms of covering traversal ground_

Pre-processing
* The investigator relies on the callgraph to be present attached to workspace context.
* We create forward callgraph then, transpose (flip the edges) it to get a reverse callgraph (for upstream navigation)
* Then attach both to Workspace Context before the pipeline is run

Trackers
* These are structures that consume from the investigators and every detector that makes use of the investigator is expected to maintain it's own tracker for tracking stuff across procedure boundaries.
* Then this information is used to communicate back to the detector what to do about the investigation

Future scope
- Inter contract analysis ? 🧐 

Add the following detectors as proof of concept to this PR

 - [x] #597 
-  [x] #600
-  [x] `x.sub(y)` without `if` checks to stop unexpected reverts (cc: @devdacian)


tests/contract-playground/src/SubWithoutIf.sol
```diff
@@ -0,0 +1,63 @@
...
+
+}
+
+contract SubtractionContract {
+    // Function to subtract two numbers
+    function sub(uint256 a, uint256 b) external pure returns (uint256) {
+        require(b <= a, "Subtraction would result in a negative number");
+        return a - b;
+    }
+}
```
TilakMaddy, line 63: This is like a mock prb

TilakMaddy:
@alexroan There are bunch of unit tests for the callgraph framework in the file `detect/experimental/cfg_sample.rs` in addition to that the 3 detectors' tests :) In fact, I want to add more but I think I want to wait for your approval :)

tests/contract-playground/src/SubWithoutIf.sol
TilakMaddy: Follow up: Should we also look for `x.div(y)` ?

aderyn_core/src/detect/low/sub_without_if.rs
```diff
@@ -0,0 +1,150 @@
...
+use eyre::Result;
+
+#[derive(Default)]
+pub struct SubWithoutIfDetector {
+    // Keys are: [0] source file name, [1] line number, [2] character location of node.
+    // Do not add items manually, use `capture!` to add nodes to this BTreeMap.
+    found_instances: BTreeMap<(String, usize, String), NodeID>,
+}
+
+impl IssueDetector for SubWithoutIfDetector {
```
alexroan, line 27: I don't quite understand which ticket this detector is related to. Could you link it?

tests/contract-playground/src/Tower.sol
```diff
@@ -0,0 +1,86 @@
...
+        if (!success) {
+            revert();
+        }
+    }
+
+    modifier passThroughNinthFloor3(address x) {
+        visitEighthFloor3(x);
+        _;
+    }
+
```
alexroan, line 69: So that we cover all bases, could you please include the other methods of ETH transfer in these tests?

`destinationAddress.transfer(balance);`
`destinationAddress.send(balance);`

alexroan (changes requested):
Could you reference all the tickets being addressed in this one please @TilakMaddy ?

tests/contract-playground/src/Tower.sol
TilakMaddy: Surely I can do this :) I just have to complete the helper function .

TilakMaddy:
@alexroan You are right, I was focussed on the callgraph library and I didn't even realize I used the same `Tower.sol` file for testing the detectors as well as the callgraph. I have now seperated them and also under relevant file names.

`SendEtherNoChecks.sol` and `CallGraphTests.sol`

aderyn_core/src/detect/low/sub_without_if.rs
TilakMaddy: Check your messages

tests/contract-playground/src/Tower.sol
TilakMaddy: ✅

aderyn_core/src/context/graph/workspace_callgraph.rs
```diff
@@ -0,0 +1,134 @@
+use std::collections::{hash_map, HashMap, HashSet};
```
alexroan, line 1: This is a big chunk of potentially _very_ impactful code. Could you please add some unit tests for each of the functions here to ensure the callgraph is being created correctly?

aderyn_core/src/context/investigator/standard.rs
```diff
@@ -0,0 +1,372 @@
+//! This module helps with strategies on performing different types of investigations.
```
alexroan, line 1: Unit tests for each function here, please? :D

aderyn_core/src/detect/helpers.rs
```diff
@@ -94,3 +97,99 @@ pub fn has_msg_sender_binary_operation(function_definition: &FunctionDefinition)
...
 }
+
+// Check if an ast_node sends native eth
+// Examples:
+// ```
+// function foo() public {
+//     address(0x1).call{value: 10}("...")
+// }
+// ```
+pub fn has_calls_that_sends_native_eth(ast_node: &ASTNode) -> bool {
```
alexroan, line 108: Not a blocker for this PR, but maybe we should think about organizing this file into a module and sub-files?

It should be super easy for new devs to find and use helper functions like this. Maybe we even move them to the ast code?

Let's put a pin in that and think about the best way to structure it :)

alexroan (changes requested):
This is looking _very_ good @TilakMaddy . I have a few change requests before another review. We're getting close though, this looks powerful af.

aderyn_core/src/detect/helpers.rs
TilakMaddy: 💯 I'm with you on that

aderyn_core/src/context/investigator/callgraph_tests.rs
TilakMaddy, line 1: The tests are here. !!!

alexroan:
I'm removing "Sub without If" from this so we can merge the callgraph without it blocking.

aderyn_core/src/detect/test_utils/load_source_unit.rs
```diff
@@ -5,13 +5,27 @@ use std::{
...
+use crate::{
+    ast::SourceUnit,
+    context::{graph::traits::Transpose, workspace_context::WorkspaceContext},
+};
+use crate::{context::graph::WorkspaceCallGraph, visitor::ast_visitor::Node};
 
 use super::ensure_valid_solidity_file;
 
+#[cfg(test)]
+pub fn load_solidity_source_unit_with_callgraphs(filepath: &str) -> WorkspaceContext {
```
alexroan, line 17: As a follow up - we should probably just remove this option so that callgraphs are loaded for every test harness. More choices/options are just more DevRel questions that we can solve by including all the bells and whistles in one load_solidity_source function.

alexroan (approved):
Approved, with a follow up that I'll make a note of @TilakMaddy

PR #607, opened 2024-07-22
Detector: Misused boolean

TilakMaddy:


aderyn_core/src/detect/high/misused_boolean.rs
```diff
@@ -0,0 +1,130 @@
...
+    // Do not add items manually, use `capture!` to add nodes to this BTreeMap.
+    found_instances: BTreeMap<(String, usize, String), NodeID>,
+}
+
+impl IssueDetector for MisusedBooleanDetector {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
+        // PLAN -
+        // Check for
+        // 1. if (… || true)`
+        // 2. if(.. && false) (on either side of the || binary operations)
```
TilakMaddy, line 26: Main logic

aderyn_core/src/detect/high/misused_boolean.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    severity: High,
+    title: "Misused boolean with logical operators",
+    desc: "The patterns `if (… || true)` and `if (.. && false)` will always evaluate to true and false respectively.",
+    name: MisusedBoolean,
+
+    |context| {
+        for binary_operation in context.binary_operations() {
+            if binary_operation.operator == "||"
+                && [
```
TilakMaddy, line 16: For simple detectors, we can use this format : )

tests/contract-playground/src/MisusedBoolean.sol
```diff
@@ -0,0 +1,37 @@
...
+        }
+        return 0;
+    }
+
+      function misuse4(uint256 num) external returns(uint256) {
+        if (true || isEven(num)) {
+            return num * num;
+        }
+        return 0;
+    }
```
alexroan, line 35: What about this case, [from here](https://github.com/crytic/slither/wiki/Detector-Documentation#misuse-of-a-boolean-constant)?

```sol

if (false) { // bad

```

(and equivalent `true` case)

aderyn_core/src/detect/mod.rs
```diff
@@ -24,4 +24,74 @@ macro_rules! capture {
...
+            found_instances: std::collections::BTreeMap<(String, usize, String), NodeID>,
+        }
+
+        impl $crate::detect::detector::IssueDetector for $detector_struct {
+
+            fn detect(&mut self, context: &$crate::context::workspace_context::WorkspaceContext) -> Result<bool, Box<dyn std::error::Error>> {
+
+                let $context = context;
+
+                macro_rules! grab {
```
alexroan, line 51: Why `grab` and not `capture`?

TilakMaddy: [Macro Hygiene Rule](https://veykril.github.io/tlborm/syntax-extensions/hygiene.html) in Rust makes it such that,  inside a macro, an externally defined macro can't make a reference to "self". Here, `capture!()` is defined external to the `issue_detector!` so it won't be able to refer to it. That makes it "unhygenic". Therefore, had to define one inside :)

TilakMaddy:
I think we need to have some kind of `unveil()` method on our AST Nodes

Right now, there is no way to express the following without manually covering every case each time

> "a constant boolean"

When I say that, I want to match with 
* Expression::Literal whose value is false/true
* VariableDeclaration that is boolean type and constant 
* Expression::UnaryOperation with ! operator followed by a sub expression that could be either of the above

alexroan:
> I think we need to have some kind of `reveal()` method on our Expression nodes

Is there a way we can do this with a `helper` or `browser` module @TilakMaddy ?

TilakMaddy:
> > I think we need to have some kind of `reveal()` method on our Expression nodes
> 
> Is there a way we can do this with a `helper` or `browser` module @TilakMaddy ?

Added to helper

TilakMaddy:
WAAAIIIT ! This workflow was running for 1hr 15 mins w/o my knowledge and I had to manually stop it ? That's bad. We need a timeout

alexroan (approved):
Approved!

PR #612, opened 2024-07-23
Detector: Unchecked return value

TilakMaddy:


aderyn_core/src/detect/detector.rs
```diff
@@ -37,6 +37,8 @@ use std::{
     str::FromStr,
 };
 
+use super::high::UncheckedReturnDetector;
```
alexroan: This may be your IDE auto-importing here. We should add these `use` statements to the large ones above ^. I've fixed in my dev merge commit :)

tests/contract-playground/src/UncheckedReturn.sol
```diff
@@ -0,0 +1,56 @@
...
+        // BAD (we're doing nothing)
+        UncheckedHelperExternal(address(0x12345)).two();
+    }
+
+    function callTwoAndDoSomething() internal pure {
+        // GOOD (we're storing the return value in a variable)
+        uint256 _answer = UncheckedHelperExternal(address(0x12345)).two();
+    }
+
+    function callTwoAndRequireSomething() internal pure {
```
alexroan, line 35: I've added these two test cases that use the return value directly in a `require` and an `if` conditional.

aderyn_core/src/detect/high/unchecked_return.rs
```diff
@@ -0,0 +1,116 @@
...
+                    referenced_declaration: Some(id),
+                    ..
+                })
+                | Expression::MemberAccess(MemberAccess {
+                    referenced_declaration: Some(id),
+                    ..
+                }) => {
+                    if function_call
+                        .parent(context)
+                        .is_some_and(|node| node.node_type() == NodeType::Block)
```
alexroan, line 42: This is great! My first instinct would be to try to find all the possible positive cases, but you're using the sole negative case here! Great stuff 👍

alexroan (approved):

PR #615, opened 2024-07-24
Release binary optimized for speed 

TilakMaddy:
Reference: https://nnethercote.github.io/perf-book/build-configuration.html

(The Rust Performance Book)


alexroan:
@TilakMaddy I get this when I run `cargo build --release`

```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/alexroan/Documents/Cyfrin/aderyn/aderyn/Cargo.toml
workspace: /Users/alexroan/Documents/Cyfrin/aderyn/Cargo.toml
```

TilakMaddy:
Hmm .. I should've added it to Cargo.toml in the root file instead of the `aderyn/Cargo.toml` Now, I don't get the warning.

alexroan (approved):

PR #616, opened 2024-07-25
Detector: Deletion of nested mapping

TilakMaddy:


tests/contract-playground/src/DeletionNestedMappingStructureContract.sol
```diff
@@ -0,0 +1,20 @@
...
+        uint256[] names;
+        mapping(address => uint256) age;
+    }
+
+    mapping(address => Person) private people;
+
+    function remove() internal{
+        // We are deleting from a mapping whose value is a struct which contains a member of type mapping.
+        // Therefore, capture it!
+        delete people[msg.sender];
```
TilakMaddy, line 17: @alexroan Is my understanding here correct? If so, this PR is ready to review!

aderyn_core/src/detect/high/deletion_nested_mapping.rs
```diff
@@ -0,0 +1,135 @@
...
+    fn severity(&self) -> IssueSeverity {
+        IssueSeverity::High
+    }
+
+    fn title(&self) -> String {
+        String::from("Deletion from a nested mappping.")
+    }
+
+    fn description(&self) -> String {
+        String::from("A deletion in a structure containing a mapping will not delete the mapping. The remaining data may be used to compromise the contract.")
```
TilakMaddy, line 89: "as-is" from Slither for parity

TilakMaddy: Can you help me understand the recommendation? I didn't quite get it : (

alexroan: Good question @TilakMaddy , it's not clear. My first thought was to just add a boolean `locked` field in the struct that you can check when retrieving the balances, but that would not prevent direct mapping references.

I'm not actually dure

alexroan (dismissed):

alexroan (approved):

PR #617, opened 2024-07-25
Detector: Usage of pre-declared variables.

TilakMaddy:
.


aderyn_core/src/detect/high/pre_declared_variable_usage.rs
```diff
@@ -0,0 +1,133 @@
...
+                if let Some(id) = used.referenced_declaration {
+                    if let Some(ASTNode::VariableDeclaration(variable_declaration)) =
+                        context.nodes.get(&id)
+                    {
+                        if !used
+                            .appears_after(context, variable_declaration)
+                            .is_some_and(identity)
+                        {
+                            capture!(self, context, used);
+                        }
```
TilakMaddy, line 68: Apparently, context.get_source_unit_for_child(node) returns None for these 2 nodes ! So we panic fail

alexroan:
I've retroactively fitted `#[serial]` to many detector tests, hoping to resolve the flakey fails we're seeing in CI @TilakMaddy.

PR #618, opened 2024-07-26
Detector: Weak Randomness

### LINKED ISSUE #326: High detector: weak randomness
The `keccak256` hash of a combination of predictable values like `block.timestamp`, `block.number`, or any values, should not be accepted as random. Relying on it could harm the protocol. Rather we should use something like Chainlink VRF which gives access to mathematically proven random values on chain.

DavidDrob:
Closes #326 

The detector checks for:
- any use of `block.prevrandao`
- modulo operations on predictable values like `block.timestamp`, `block.number` and `blockhash`
- hashing those values using `keccak256`

The test cases are based on those by slither and two solodit issues.
- https://github.com/crytic/slither/blob/master/tests/e2e/detectors/test_data/weak-prng/0.7.6/bad_prng.sol
- https://solodit.xyz/issues/m-06-bad-source-of-randomness-code4rena-holograph-holograph-contest-git
- https://solodit.xyz/issues/m-04-protocol-uses-manipulatable-randomness-pashov-none-subsquid-markdown


TilakMaddy:
Looks good man ! 

Can you please do the following ?

a) Run `cli/reportgen.sh` on this
b) Remove the `unwrap()` stuff

aderyn_core/src/detect/high/weak_randomness.rs
```diff
@@ -0,0 +1,190 @@
...
+                    capture!(self, context, keccak);
+                }
+            }
+            // get variable definition
+            else if let Expression::Identifier(ref i) = *arg {
+                if let Some(node_id) = i.referenced_declaration {
+                    let decleration = context.get_parent(node_id);
+
+                    if let Some(ASTNode::VariableDeclarationStatement(var)) = decleration {
+                        if let Some(Expression::FunctionCall(function_call)) = &var.initial_value {
```
TilakMaddy, line 43: Can you please explain why use `get_parent` ? If we want to get the variable declaration we can do `context.nodes.get(node_id)` right?

DavidDrob: I need `VariableDeclarationStatement` because it already includes the `initial_value`. If I used `context.nodes.get(node_id)` I'd also need to get its parent, I think, resulting in more code.

But please tell me if you know an easier way to get the `initial_value`.

DavidDrob:
> Looks good man !
> 
> Can you please do the following ?
> 
> a) `cargo clippy -- -D warnings` b) Run `cli/reportgen.sh` on this c) Remove the `unwrap()` stuff

Done :heavy_check_mark: 

Thanks for the feedback, let me know if there's anything else I should update. I left the variable declaration part as it was, because I couldn't find a simple way to get `initial_value` using just `VariableDeclaration`.

TilakMaddy:
Thank you for taking interest @DavidDrob !

TilakMaddy:
@DavidDrob You have to run `make` . Then run `cli/reportgen.sh` because right now, there is no reported generated for templedao. In order to do that we have to install the dependencies and make is the tool that will help you : )

alexroan (approved):
@DavidDrob Congratulations on your first contribution!

This detector will be released in v0.1.8! 🥳

PR #620, opened 2024-07-27
Unimplemented functions detector [Help needed]

TilakMaddy:
Thoughts: This is getting tricky. 

Scope:
For Solidity 0.5 and below
(You get errors in the new versions so no need to show)

Current Logic:
1. Target a contract.
2. Check to see if there is a implemented function for every unimplemented function definition in the linearized base contracts chain.

3. So far, easy. But there is an edge case. We can have public variables that can act as functions and override unimplemented function definitions in the hierarchy. This makes it tricky to find because of the below limitation

Field limitations of using AST:
> We don't get to check for `internalFunctionIDs` and get the function selectors on contracts (for older solc versions like 0.5). So I wrote my own logic for deriving something *like* a function signature - called it `function signaturish`.
As a result, was able to detect all the unimplemented functions in the inheritance hierarchy that were  implemented by another function. 

What I'm unable to figure out is how might we deal with detecting public state variables that override functions .


reports/report.md
TilakMaddy, line 1: Hang on, why did we lose everything ?

PR #629, opened 2024-07-29
Fix title & desc - state variable shadowing

TilakMaddy:
cc : @chatgpt 🤣


TilakMaddy:
Since, audit wizard is using this, it **may** be a case that we have to run another release  (But on the other hand, it's a minor one)

aderyn_core/src/detect/high/state_variable_shadowing.rs
```diff
@@ -179,11 +179,19 @@ impl IssueDetector for StateVariableShadowingDetector {
...
 
     fn title(&self) -> String {
-        String::from("High Issue Title")
+        String::from("Shadowed State Variables in Inheritance Hierarchy")
     }
 
     fn description(&self) -> String {
-        String::from("Description of the high issue.")
+        String::from(
+            "This vulnerability occurs when a state variable in a derived contract unintentionally\
```
alexroan, line 187: Can you make this more succinct, please, @TilakMaddy ? 2 or 3 sentences maximum. Remember that it should be easily readable within a vscode diagnostic popup.

alexroan (changes requested):

aderyn_core/src/detect/high/state_variable_shadowing.rs
TilakMaddy: ✅

alexroan (approved):

PR #630, opened 2024-07-29
Detector: Contract that locks ether

TilakMaddy:


tests/contract-playground/src/ContractLocksEther.sol
```diff
@@ -0,0 +1,137 @@
+// SPDX-License-Identifier: MIT
```
alexroan, line 1: This is a great test file that covers inheritance, but I don't see any reference to eth `send()` and `transfer()` functions that could also act as a withdraw function. Could you please add cases for these?

alexroan (changes requested):
@TilakMaddy I like how you approached this one with regards to the detector logic structure.

I do have one request to improve coverage.

tests/contract-playground/src/ContractLocksEther.sol
TilakMaddy: ✅

alexroan (dismissed):

alexroan (approved):

PR #631, opened 2024-07-29
Detector: Incorrect Modifiers + Discord Issue Fix

### LINKED ISSUE #632: Running `aderyn .` fails with `data did not match any variant of untagged enum BlockOrStatement`
Repo:
https://github.com/code-423n4/2024-07-munchables

Source:
Discord

TilakMaddy:
Detector. :) 

Others
-------

Fixes #632 
```
`Running `aderyn .` fails with `data did not match any variant of untagged enum BlockOrStatement` issue in Discord
```


aderyn_core/src/ast/ast_nodes.rs
```diff
@@ -639,7 +639,7 @@ ast_node!(
 ast_node!(
     #[derive(Hash)]
     struct Return {
-        function_return_parameters: NodeID,
+        function_return_parameters: Option<NodeID>,
```
TilakMaddy, line 642: This should have been optional. Because  it wasn't, it has affect at least the contracts where modifier has a return statement.

FIX

reports/report.md
```diff
@@ -1672,13 +1674,60 @@ A deletion in a structure containing a mapping will not delete the mapping. The
...
+	    modifier whenNotPaused() {
+	```
+
+- Found in src/IncorrectModifier.sol [Line: 46](../tests/contract-playground/src/IncorrectModifier.sol#L46)
+
+	```solidity
+	    modifier whenNotPaused2() {
+	```
+
+- Found in src/OnceModifierExample.sol [Line: 6](../tests/contract-playground/src/OnceModifierExample.sol#L6)
```
alexroan, line 1702: Hmmm, This is just an empty modifier, but I don't think this detector should catch this one.

reports/report.md
```diff
@@ -1672,13 +1674,60 @@ A deletion in a structure containing a mapping will not delete the mapping. The
...
+
+	```solidity
+	    modifier onlyOnce() {
+	```
+
+- Found in src/SendEtherNoChecks.sol [Line: 70](../tests/contract-playground/src/SendEtherNoChecks.sol#L70)
+
```
alexroan, line 1709: This one definitely shouldn't be caught

alexroan (changes requested):
Some false positives to fix @TilakMaddy

TilakMaddy:
Noted. I will do the following then .. 

* Make sure modifiers in abstract contracts are not caught
* If there is a high level call in the modifier, or a transfer to an address, don't catch the detector (because we will assume they will take care of reverting)

reports/report.md
TilakMaddy: For this one, I think it is should be caught because if a user disables the empty block detector, doesn't automatically mean they are okay with having a modifier that doesn't potentially revert when there is an undesired transaction. Also the empty detector is LOW, this one is HIGH

reports/report.md
```diff
@@ -1672,13 +1674,60 @@ A deletion in a structure containing a mapping will not delete the mapping. The
...
+	```solidity
+	    modifier mod1(address x) {
+	```
+
+- Found in src/WrongOrderOfLayout.sol [Line: 17](../tests/contract-playground/src/WrongOrderOfLayout.sol#L17)
+
+	```solidity
+	    modifier isZero(uint256 value) {
```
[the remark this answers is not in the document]

reports/report.md

reports/report.md
TilakMaddy: Hmm I disagree on this one because looking into this description in Slither's Wiki, I am convinced we shouldn't have a modifier that just has a placeholder staement.

> If a modifier does not execute _ or revert, the execution of the function will return the default value, which can be misleading for the caller.

alexroan: I agree that placeholder modifiers are bad practice, but this is outside the scope of this detector. It should be a separate detector issue.

This detector is specifically intended to catch modifiers with a path inside it that does not include the `_;`, which this modifier does not adhere to.

alexroan (changes requested):

TilakMaddy:
Umm I'm still not clear. 

So, the current criteria for good modifier in this detector is that 

placeholder should be present 
        AND 
(revert OR require should be present)


Now, as per the requested changes,  we want the criteria for a good detector to be just 

placeholder should be present

Is that right ??

alexroan:
Criteria:

If there are any modifiers with an execution path that doesn't lead anywhere, we capture it.

From the developer's perspective, every execution path within a modifier should have either
* `_`
* revert OR require OR address.transfer()

Does that make sense?

TilakMaddy:
@alexroan Now I see ..... So 

`_` OR revert OR require OR address.transfer()

alexroan:
Yes, every path must have at least one of those things.

TilakMaddy:
Converting to draft because it's much more nuanced than that.

It's not enough to just have a `_` in the modifier. Upon reading the detector again I find that the real requirement (and as you also have stated) is that there must be an execution path for all cases in a given modifier. This means if we look at the exploit scenario described for this detector, it's clear even though there is a `_;` , it's inside an if statement. So the problem is when the if condition fails, the function code is not executed. 

Looks like we need a control flow graph to be able to write such type of rules. So let me draft this and remove it from review.
I will also send a separate PR for fixing #632  right now

PR #636, opened 2024-07-31
Detector: `msg.value` used in loop

TilakMaddy:


aderyn_core/src/detect/high/msg_value_in_loops.rs
```diff
@@ -0,0 +1,140 @@
...
+        IssueSeverity::High
+    }
+
+    fn title(&self) -> String {
+        String::from("Loop contains `msg.value`.")
+    }
+
+    fn description(&self) -> String {
+        String::from("Do not use `msg.value` in loops since it may be misleading when calculating the total amount.\
+            It's recommended to prepare additional variables outside the loop to do the required calculation.")
```
alexroan, line 62: Let's be more explicit here. Taken from the Slither wiki:

"Provide an explicit array of amounts alongside the receivers array, and check that the sum of all amounts matches msg.value."

tests/contract-playground/src/MsgValueInLoop.sol
```diff
@@ -0,0 +1,63 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
```
alexroan, line 3: Can you please add `// GOOD` examples, written using the recommendation we provide?

alexroan (changes requested):

tests/contract-playground/src/MsgValueInLoop.sol
TilakMaddy: Of course !

aderyn_core/src/detect/high/msg_value_in_loops.rs
TilakMaddy: Hmm this one I thought was too specific of a description. But I think explicit is good. We'll do that :)

PR #638, opened 2024-07-31
Detector: High level calls inside a loop

TilakMaddy:
Pre-req #637


aderyn_core/src/detect/high/high_level_calls_in_loop.rs
```diff
@@ -0,0 +1,195 @@
...
+    fn severity(&self) -> IssueSeverity {
+        IssueSeverity::High
+    }
+
+    fn title(&self) -> String {
+        String::from("Loop contains high level calls to other contracts and/or eth is transferred.")
+    }
+
+    fn description(&self) -> String {
+        String::from("One of the dangers of having high level calls or an `address.transfer()` inside a loop is that \
```
alexroan, line 84: This sounds very different to issue ADE-54 in linear. The recommendation there is to use pull over push because it prevents DDOS attacks. Here we're saying the issue is something else. Can you explain this some more please?

alexroan (changes requested):

aderyn_core/src/detect/high/high_level_calls_in_loop.rs
```diff
@@ -0,0 +1,195 @@
...
+    2. Contract.xyz()
+
+    Things what we don't capture - `address.send`, `address.call`, etc
+    (Because we know for sure they don't revert unless we ask them to based on the return value of bool success)
+
+    The solution is to perform this function call with low level
+    call and then you can just break off the loop if the bool success returns false!
+
+    In order to catch `requires` and `reverts` in the loops, there is another detector we have in `reverts_and_requires_in_loop.rs`
+*/
```
TilakMaddy, line 39: @alexroan It's explained here. 

But I'll change the description to match slither's :)

TilakMaddy: TBH I'm unsure of this detector and this was the one for which you said you'll have a think about whether or not to consider because pull over  push is more of an archaic concept now

TilakMaddy: I'm totally okay with not proceeding with this. What do you say?

aderyn_core/src/detect/high/high_level_calls_in_loop.rs
TilakMaddy: Changed to match with slither. 
✅

aderyn_core/src/detect/high/high_level_calls_in_loop.rs
alexroan: I'm thinking about it, I'm not ready to merge this yet based on this discussion.

Can you please:
1. Link this PR on the ticket
2. Move the ticket back to the backlog
3. Close this PR

We can reassess whether we want to tackle this one at the sync meeting.

TilakMaddy: Yessir!

PR #639, opened 2024-07-31
Feature/extract manipulated state variables

TilakMaddy:


aderyn_core/src/context/browser/extractor.rs
```diff
@@ -125,3 +127,179 @@ impl ASTConstVisitor for ExtractReferencedDeclarations {
...
+
+        let mut total_state_variables_assigned = 0;
+
+        for contract in context.contract_definitions() {
+            let state_variables_info = ExtractManipulatedStateVariablesIDs::from(contract);
+            println!(
+                "{} - {}",
+                contract.name,
+                state_variables_info.assigned.len()
+            );
```
TilakMaddy, line 298: We can use this pattern :)

TilakMaddy:
oops I forgot something really basic. I'm drafting this 😬

PR #641, opened 2024-08-02
Detector: constant functions contains assembly

TilakMaddy:


aderyn_core/src/detect/low/constant_funcs_assembly.rs
```diff
@@ -0,0 +1,131 @@
...
+#[derive(Default)]
+pub struct ConstantFunctionContainsAssemblyDetector {
+    // Keys are: [0] source file name, [1] line number, [2] character location of node.
+    // Do not add items manually, use `capture!` to add nodes to this BTreeMap.
+    found_instances: BTreeMap<(String, usize, String), NodeID>,
+}
+
+impl IssueDetector for ConstantFunctionContainsAssemblyDetector {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
+        for function in helpers::get_implemented_external_and_public_functions(context) {
```
alexroan, line 28: Why only public and external?

aderyn_core/src/detect/low/constant_funcs_assembly.rs
```diff
@@ -0,0 +1,131 @@
...
+            if function.state_mutability() == &StateMutability::View
```
alexroan, line 29: This issue only presents itself if the pure/view function that contains the assembly can be compiled with solce <0.5. We should filter for this also, otherwise we'll get lots of false positives for valid assembly within pure functions in 0.8.0, for example.

alexroan (changes requested):

aderyn_core/src/detect/low/constant_funcs_assembly.rs
TilakMaddy: If an internal function can never be invoked  through any public / external function has assembly, it's okay :)

TilakMaddy: I use StandardInvestigator which will visit those  internal and private functions that can be visited through these public/external functions

aderyn_core/src/detect/low/constant_funcs_assembly.rs
TilakMaddy: ✅

reports/templegold-report.md
TilakMaddy, line 1: This stuff has to be investigated I think.. what's happened here?

alexroan (approved):

PR #643, opened 2024-08-02
Detector: Unused `private` / `internal` state variables

TilakMaddy:


reports/templegold-report.md
```diff
@@ -8653,3 +8654,26 @@ Remove the redundant statements because no code will be generated and it just co
...
+State variable appears to be unused. No analysis has been performed to see if any inilne assembly references it. So if that's not the case, consider removing this unused variable.
+
+<details><summary>2 Found Instances</summary>
+
+
+- Found in contracts/fakes/templegold/TempleGoldStakingMock.sol [Line: 71](../tests/2024-07-templegold/protocol/contracts/fakes/templegold/TempleGoldStakingMock.sol#L71)
+
+	```solidity
+	    mapping(address delegate => uint256 balance) private _delegateBalances;
+	```
```
TilakMaddy, line 8668: Should we exclude variable names starting with `_` ??

TilakMaddy: But I think no, because this is inside a `contract/fakes/...` So ideally when you run aderyn you would exclude fakes directory I'm guessing

alexroan: This should be included! This is good. It's up to the developer whether they want to exclude the `fakes/` directory!

alexroan (approved):

PR #647, opened 2024-08-03
Check Github releases for upgraded version, instead of crates.io

alexroan:
* Check new version tags from Github releases, rather than crates.io.
* Add a warning in the crate README.

### DIFF
--- a/aderyn/README.md
+++ b/aderyn/README.md
@@ -1,4 +1,9 @@
 
+> ⚠️ **Installing via crates is no longer fully supported. `cyfrinup` is the preferred installation method.**.
+>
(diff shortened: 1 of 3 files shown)


TilakMaddy:
This looks good. But for existing cargo installers, do we create a separate branch just for 1 release where we tell them about `cyfrinup` ?

alexroan:
Good idea. in a follow up branch:
1. create a big warning after every run to use `cyfrinup` instead of cargo.
2. publish the package to crates.io
3. yank old versions from crates.io.

Can you approve this one so I can merge?

TilakMaddy (approved):
Yes please !! @alexroan

PR #648, opened 2024-08-04
Detector: Local variable shadowing

TilakMaddy:
Severity: Low

Detects non state variables in non abstract contracts whose names exactly match with *any* of state variables in *that* contract's heirarchy.

Exception: When the local variable is part of event definition or error definition, we don't capture it!

So, for example we don't catch this variable definition of roll - `event RollChanged(uint roll)` even though there is a state variable called `roll` somewhere in the contract or it's linearized base chain heirarchy


aderyn_core/src/detect/low/local_variable_shadowing.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        Ok(!self.found_instances.is_empty())
+    }
+
+    fn severity(&self) -> IssueSeverity {
+        IssueSeverity::Low
+    }
+
+    fn title(&self) -> String {
+        String::from("Local variable shadows state variables in the contract hirearchy")
```
alexroan, line 68: Slight title grammar nit that I fixed @TilakMaddy  (removed "it's")

alexroan (changes requested):
This is somehow non-deterministic.

If you keep running ./cli/reportgen.sh locally, you sometimes get different results.

Can you investigate @TilakMaddy ?

TilakMaddy:
Wait ... for some reason I see 
>NEW VERSION OF ADERYN AVAILABLE! Please run `cargo install aderyn` to fully upgrade the current version

when I run cli/reportgen

TilakMaddy:
Hmm I tried 3 times, but I got the same results. But it was evident from CI that the report missed like an instance or 2. And when you run it again, it goes away. Not sure.

reports/templegold-report.md
```diff
@@ -6783,12 +6783,6 @@ Use descriptive reason strings or custom errors for revert paths.
 	        require(address(pools[revalToken]) == address(0));
 	```
 
-- Found in contracts/util/ABDKMath64x64.sol [Line: 36](../tests/2024-07-templegold/protocol/contracts/util/ABDKMath64x64.sol#L36)
```
alexroan, line 6786: This is blocked until we understand why this is happening @TilakMaddy

TilakMaddy: Could this be hash collisions of sorts? In the workspace context, we have things like `assignments_context`, `returns_context`, etc... which are hashmaps from keys being the hash of the node and values being `NodeContext` ? 

This way, at the time of capturing, due to hash collision, we get the wrong value, here source unit id ??

Not sure, just speculation

alexroan:
This is currently blocked until non-determinism is resolved. Closing due to inactivity.

PR #649, opened 2024-08-04
Feature: Find State Variable manipulation instances

TilakMaddy:
Lightweight tool for approximately finding manipulated state variables.

Name: `ApproximateStorageChangeFinder`

Given an AST Block, it tries to detect any state variable inside it that may have been manipulated.
Now it's important to know that manipulations can occur either directly by assigning to a state variable
or, it may occur by assigning to a storage pointer that points to some state variable.
This light weight state variable manipulation finder captures both of the above kinds of assignments.
However, it's not smart enough to use a data dependency graph to determine the exact state variables
these storage pointers would be pointing to, in the context of the block-flow. PS: If it's a 1 level storage pointer 
assignment, it will be able to catch that. Have no expectation for it to work when there are multiple levels of storage pointers or they are being passed around as return values from internal function / packed in tuple expressions, etc..

NOTE - Assignment is not the only avenue for manipulating state variables, but also operations like
`push()` and `pop()` on arrays, `M[i] = x` on mappings, `delete X` imply the same.

Here, this tool covers all those kinds of changes.

Usage/Interface:

```rust
let finder_results = ApproximateStorageChangeFinder::from(ast_node);

finder_results.state_variables_have_been_manipulated() 
finder_results.no_state_variable_has_been_manipulated()
finder_results.state_variable_has_been_manipulated(&variable_declaration)
finder_results.state_variable_has_not_been_manipulated(&variable_declaration)

// The above functions return Option<bool> So if it's None that means this finder has no opinion on whether or not it is true/false

// Heuristics have been used to make it such that it returns Some more frequently than None thereby making it useful for detectors

finder_results.fetch_non_exhaustive_manipulated_state_variables()

// This above one says it in it's name. So we'll take it for what it is.

```

Thoughts - I am not super-impressed but it's far from terrible I guess. There's definitely scope for improvement. There are a lot of cases that need to be covered especially when things are being unpacked from tuples or returned from function call chains, etc, etc.

Dependency graph would be great to have in the future :) 

I had also spend a. lot of time trying to track storage "reads" but removed it from the commits. It was terrible.


tests/contract-playground/src/StateVariablesManipulation.sol
TilakMaddy, line 1: LITERALLY every line here is tested

TilakMaddy:
I'm waiting to use it in a detector before opening this for review.

aderyn_core/src/context/browser/finders.rs
TilakMaddy, line 1: Main file

TilakMaddy:
Verify

+=
-=

alexroan:
I am closing this in favor of the follow-up PRs with the state-variable-change tag

Please reopen this one if I'm mistaken @TilakMaddy

TilakMaddy:
> I am closing this in favor of the follow-up PRs with the state-variable-change tag
> 
> Please reopen this one if I'm mistaken @TilakMaddy 

Yes that's fine. I haven't pushed any new changes to this branch after using this on the detectors....

PR #650, opened 2024-08-04
Feature: Add hints per instance found

TilakMaddy:
With this PR, you can optionally insert a 4th parameter in `capture!(...)` which is a String that is specifically used to give hints/describe the issue related to that instance only.

PoC:

For the contracts with todos detector, so far we've captured the contract but now we also have a hint associated to each contract instance captured, which can guide the user to the exact line where the TODO is because that's been included in the hint


reports/report.json
```diff
@@ -3748,13 +3748,15 @@
...
+            "src_char": "66:8",
+            "hint": "A TODO comment was found on line 5"
           },
           {
             "contract_path": "src/Counter.sol",
             "line_no": 4,
             "src": "74:7",
-            "src_char": "74:7"
+            "src_char": "74:7",
+            "hint": "A TODO comment was found on line 12"
```
TilakMaddy, line 3759: extra "hint" field in JSON here

reports/report.md
```diff
@@ -3810,12 +3810,14 @@ Contract contains comments with TODOS
...
 - Found in src/ContractWithTodo.sol [Line: 4](../tests/contract-playground/src/ContractWithTodo.sol#L4)
 
+	A TODO comment was found on line 5
 	```solidity
 	contract TodoList {
 	```
 
 - Found in src/Counter.sol [Line: 4](../tests/contract-playground/src/Counter.sol#L4)
 
+	A TODO comment was found on line 12
```
TilakMaddy, line 3820: Hint appears on the line just above the code preview

reports/report.sarif
TilakMaddy, line 1: (Working on sarif support right now)

aderyn_core/src/detect/low/contracts_with_todos.rs
```diff
@@ -33,7 +34,21 @@ impl IssueDetector for ContractsWithTodosDetector {
...
+                                        )
+                                    } else {
+                                        format!(
+                                            "A TODO comment was found on line {}",
+                                            token.start_line
+                                        )
+                                    }
+                                };
+
+                                capture!(self, context, contract, hint);
```
TilakMaddy, line 51: Like this !!

TilakMaddy:
Sorry fat finger  button presses (~Ready for review~ )

TilakMaddy:
@alexroan I was referring to this one

aderyn_core/src/detect/low/_template.rs
```diff
@@ -22,6 +22,7 @@ pub struct TemplateDetector {
     // Keys are: [0] source file name, [1] line number, [2] character location of node.
     // Do not add items manually, use `capture!` to add nodes to this BTreeMap.
     found_instances: BTreeMap<(String, usize, String), NodeID>,
```
alexroan, line 24: I don't _love_ doubling up everywhere. Seems prone to mistakes / missing things.

Can we consolidate these into one?

TilakMaddy: I agree. Let me try maintaining a hints dictionary in the detector structs (key -> error source location , value -> hints)

alexroan: This has not been resolved @TilakMaddy ?

TilakMaddy: Sorry for leaving it hanging ... Moved to draft

TilakMaddy: ✅

aderyn_core/src/detect/high/_template.rs
```diff
@@ -22,6 +22,7 @@ pub struct TemplateDetector {
     // Keys are: [0] source file name, [1] line number, [2] character location of node.
     // Do not add items manually, use `capture!` to add nodes to this BTreeMap.
     found_instances: BTreeMap<(String, usize, String), NodeID>,
+    hints: BTreeMap<(String, usize, String), String>,
 }
```
TilakMaddy, line 26: Key is the same as found_instances but the value refers to to the actual hint associated to the instance

TilakMaddy:
This is going to be so cool . We can actually add explanation to soo many issues now !! It's going to make it a lot less confusing for our users <3

reports/report.md
```diff
@@ -4670,12 +4670,14 @@ Contract contains comments with TODOS
...
 - Found in src/ContractWithTodo.sol [Line: 4](../tests/contract-playground/src/ContractWithTodo.sol#L4)
 
+	A TODO comment was found on line 8
 	```solidity
 	contract TodoList {
 	```
 
 - Found in src/Counter.sol [Line: 4](../tests/contract-playground/src/Counter.sol#L4)
 
+	A TODO comment was found on line 15
```
alexroan, line 4680: @TilakMaddy , this still doesn't feel quite right to me.

1. The `ContractDefinition` is captured in this detector, which is on line 4
2. This "hint" suggests line 15 (confusing for a developer reading both lines 4 and 15)
3. The Counter contract actually has 3 instances of `// TODO`, line 15, 20 and 22. This is not captured in hints.

IMO, simply altering this detector to detect the actual lines that `// TODO` is on would be preferable over adding this "hints" feature in its current form.

alexroan (changes requested):
I'm not quite convinced this feature add much value add atm @TilakMaddy . 

Why is this preferable over simply making the detector `detect` function more precise?

reports/report.md
TilakMaddy: > IMO, simply altering this detector to detect the actual lines that // TODO is on would be preferable over adding this "hints" feature in its current form.

For this case, yes I agree ^^ 

But I think I'll make a better case with some other detector as example. 
For instance:
When we capture the for loop, it has a problematic function call somewhere deep in the downward callgraph we can specify that as a `hint`

I think without, that there is too much work user has to do find exactly why the for loop was captured.


Coming back to this case, only line 15 captured, but after that correction is made, then line 20 will be shown.
So it's incremental. Just like clippy. Sometimes you have correct what it shows you first, to be able to see more things that go wrong.


Shall I undo the changes for todo detector ?

What do you think about the general structure though for `hints` (irrespective of the detector example) ? @alexroan

alexroan: - Given that it highlights the captured instances, how would this feature be applied in the VSCode extension?
- Why are "hints" required instead of simply making the detect function more precise? And can you show me an example of a detector where it is?

TilakMaddy: *In VSCode, Hints would be appended to the general description, but only for that particular instance.
(Read above for more explanation)

* Okay, I will add a better example to this PR :)

TilakMaddy:
Added hints example for function selector collision.

reports/report.md
```diff
@@ -2326,12 +2326,14 @@ Function selector collides with other functions. This may cause the solidity fun
...
 	```solidity
 	    function withdraw(uint256) external {
 	```
 
 - Found in src/FunctionSignatureCollision.sol [Line: 13](../tests/contract-playground/src/FunctionSignatureCollision.sol#L13)
 
+	collides with the following function name(s) in scope: withdraw
 	```solidity
 	    function OwnerTransferV7b711143(uint256) external {
 	```
```
TilakMaddy, line 2339: See here !

TilakMaddy: @alexroan

reports/report.md
TilakMaddy: ✅
DONE

PR #651, opened 2024-08-05
Detector: Domain separator collision

TilakMaddy:
Severity: High 

(We don't have medium)

Scan for contracts that are of type ERC20 (which we determine by looking at the names of inherited contracts like we do for the unsafe mint detector where we look for ERC721 contracts)

Then, look at all the public state variables along with public/external functions and try to determine if their function signature collides with `DOMAIN_SEPARATOR()`

Full definition
> function DOMAIN_SEPARATOR() external view returns (bytes32)

If we do find any, we report them. Test cases are extensively covered in the contract playground file included with the detector


tests/contract-playground/src/DomainSeparatorCollision.sol
```diff
@@ -0,0 +1,83 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
```
alexroan, line 2: Lower this version to one that accepts two functions with the same sig:
```
withdraw(uint256)
OwnerTransferV7b711143(uint256)
```

source: https://www.reddit.com/r/ethdev/comments/6n6t9w/function_hash_collision_potential_misuse/

TilakMaddy:
Here, we check for collision with DOMAIN_SEPARATOR even though the function DOMAIN_SEPARATOR  is not defined anywhere. So it wouldn't be a sub-case when checking for any 2 func sign collisions with different names

alexroan (dismissed):
Edit: apologies, let's sync again.

The last time we spoke, we were widening the scope to do the following:

1. Collect all signatures, and names (maybe in a HashMap<sig, name>)
2. If there are more than one name to a sig, capture it.

TilakMaddy:
> Edit: apologies, let's sync again.
> 
> The last time we spoke, we were widening the scope to do the following:
> 
> 1. Collect all signatures, and names (maybe in a HashMap<sig, name>)
> 2. If there are more than one name to a sig, capture it.

No problem .. I got it  .... I was just wondering if we can have that as a separate detector because that wouldn't cover this case? .. or shall I close this PR and make that one only.

alexroan:
Closed as suggested. I've also moved the ticket back into the TODO column

PR #655, opened 2024-08-05
Detector: Incorrect ERC721 contract

TilakMaddy:
Severity: High

We look for ERC721 method signatures on contract that are deemed to be ERC721 (by looking at the names of inheritance heirarchy) 
Then, if a function signature matches, but the function itself does not return what it's supposed to return according to the ERC721 spec we capture it


aderyn_core/src/detect/high/incorrect_erc721_interface.rs
```diff
@@ -0,0 +1,319 @@
...
+                                capture!(self, context, func);
+                            }
+
+                            if (func.represents_erc721_is_approved_for_all()
+                                || func.represents_erc165_supports_interface())
+                                && !func.returns_bool()
+                            {
+                                capture!(self, context, func);
+                            }
+                        }
```
TilakMaddy, line 90: Heart of the logic

reports/report.md
```diff
@@ -1988,6 +1996,71 @@ It appears that the contract includes a payable function to accept Ether but lac
...
+- Found in src/IncorrectERC721.sol [Line: 71](../tests/contract-playground/src/IncorrectERC721.sol#L71)
+
+	```solidity
+	    function safeTransferFrom(
+	```
+
+- Found in src/IncorrectERC721.sol [Line: 103](../tests/contract-playground/src/IncorrectERC721.sol#L103)
+
+	```solidity
+	    function supportsInterface(
```
alexroan, line 2057: This is not defined in ERC721. Should it be caught @TilakMaddy ?

TilakMaddy: I had the same doubt but I saw that Slither codebase uses this ...so I decided to add it

alexroan: That's a bug IMO, this function is from ERC165. Let's only catch instances where 721 is incorrect!

alexroan (changes requested):
This detector should focus only on ERC721 interface

reports/report.md

alexroan (approved):

PR #659, opened 2024-08-08
Detector: State variable initialized by function 

TilakMaddy:
Severity: Low

Capture criteria:
When the function used for initializing a state variable outside of functions (so basically inside the contract, but outside the function) makes *any* reference to non constant state variables. 

Recommended:
Do all of that in the constructor if needed to.


reports/report.md
```diff
@@ -4519,3 +4563,32 @@ A low level callee may consume all callers gas unexpectedly. Avoid unlimited imp
...
+	    uint public x = set(); // Initialize from function (sets to 88)
+	```
+
+- Found in src/FunctionInitializingState.sol [Line: 9](../tests/contract-playground/src/FunctionInitializingState.sol#L9)
+
+	```solidity
+	    uint public f = tes();
+	```
+
+</details>
```
TilakMaddy, line 4591: Here it is. It captures the variable declarations whose value is a function call *and*  the function definition makes references to non constant state variables. 

(This works regardless if it's marked as view or not.)

alexroan (approved):

alexroan (approved):

PR #662, opened 2024-08-12
Detector: Costly operations with loops

TilakMaddy:
Severity: Low

Qualification: All solc versions

Logic:
Investigate every kind of loop (for, while, do-while, etc)
Then check to see if there is any type of state change. If there is, then it's a costly operation.


tests/contract-playground/src/CostlyOperationsInsideLoops.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.4.0;
+
+// COPIED from SLITHER Wiki
+contract CostlyOperationsInLoop {
+    uint loop_count = 100;
+    uint state_variable = 0;
+
+    function bad() external {
+        for (uint i = 0; i < loop_count; i++) {
```
alexroan, line 10: @TilakMaddy , Solidity FYI, having `loop_count` in the for statement instead of a local variable incurs an SLOAD on every loop 👀

alexroan (approved):

tests/contract-playground/src/CostlyOperationsInsideLoops.sol
TilakMaddy: Got it :)

TilakMaddy: Actually, it's a good point, we need to investigate the call graph downwards even from the conidtion and the update expression of the for loop. (body is not enough) .. to find sstores. 

(In this case it was sload so I guess it was okay)

TilakMaddy: Ah wait... this detector looks inside the entire for loop. So it should cover that case. The only false positive would be if there is a sstore in the initial expression haha .  Will resolve it as tech debt

PR #666, opened 2024-08-14
star history in readme

TilakMaddy:
Star History


alexroan:
@Eversmile12 , thoughts on this?

It adds social proof but doesn't look very polished, IMO.

PR #667, opened 2024-08-14
Simplify callgraph

alexroan:
Simplify callgraph to make only (what was previously called) "Downstream" and "Forward" available to the detector, given that that is the only useful application of it so far.


alexroan:
* TODO: Reintroduce Upstream and Sideeffects

(probably a new branch)

alexroan:
Following a discussion with @TilakMaddy, the name and directory structure changes will be applied in a separate PR.

PR #669, opened 2024-08-16
Refactor Callgraph

alexroan:
### 1. Direction Naming

* "Forward" & "Downstream" are now "Inward"
* "Backward" & "Upstream" are now "Outward".

Rationale: mimic debugging terminology. 
When developers use a debugger to navigate the call stack, they "Step Into" and "Step Out" of function calls.

### 2. "Investigator" Keyword Removed

The "Investigator" keyword doesn't quite fit here in the detector context. It feels more like the structure that I'm going to send something (currently called Trackers) into to find information.

As a Detector Developer, I want to create a CallGraph from an ASTNode, and delve inward (or outward) to check for certain conditions that are present themselves further into the stack.

### 3. Directory Structure

Moved everything related to callgraphs inside the `graph` folder.


TilakMaddy:
Looks really nice! I l liked it :) Please merge

TilakMaddy:
Follow up:
(after merging this)

I will adjust all the open PRs as well to reflect these new names

TilakMaddy (approved):

TilakMaddy:
Some changes I have made:

* There was a method called `investigate` as well which I renamed to `accept` (to stay consistent) 
* Re-export to graph namespace so it's easier for detector modules to use it
* `for_specific_nodes` -> `from_many_nodes`

TilakMaddy (approved):

PR #671, opened 2024-08-18
Detector: Uninitialized local variables (Already reviewed before)

TilakMaddy:
Brings back https://github.com/Cyfrin/aderyn/pull/646 

But this time it's possible to `forge build` (Notice - there is no `--via-ir` flag requirement)

This was possible because a function with 24 local variables has been split into 2 functions (12 each)


tests/contract-playground/src/UninitializedLocalVariables.sol
```diff
@@ -0,0 +1,74 @@
+// SPDX-License-Identifier: MITc
```
alexroan, line 1: Small typo that I will fix here @TilakMaddy

tests/contract-playground/src/UninitializedLocalVariables.sol
```diff
@@ -0,0 +1,74 @@
...
+pragma solidity ^0.8.0;
```
alexroan, line 2: I'll also change this to `0.8.19` @TilakMaddy

tests/contract-playground/src/UninitializedLocalVariables.sol
TilakMaddy: @alexroan Ahh this was `zed` editor's bug . It appended `c` whenever I used a keybinding to navigate away to another file. Thanks for correcting though.

PR #672, opened 2024-08-19
Detector: State variable that could be declared constant

TilakMaddy:
Severity: Low

Logic:
1. Collect all state variables that are not marked constant or immutable and are also not structs/mappings/contracts (collection A)
2. Investigate every function and collect all the state variables that could have changed (collection B)
3. Result = collection A - collection B


reports/report.md
```diff
@@ -2608,6 +2659,18 @@ Consider using a specific version of Solidity in your contracts instead of a wid
 	pragma solidity ^0.4.0;
 	```
 
+- Found in src/StateVariableCouldBeDeclaredConstant.sol [Line: 2](../tests/contract-playground/src/StateVariableCouldBeDeclaredConstant.sol#L2)
```
alexroan, line 2662: We should follow the advice that aderyn gives. We should avoid writing new contracts with known issues that are captured by other detectors, unless it's unavoidable.

* For example, giving the contract a specific pragma can avoid this open pragma issue.
* An example of an unavoidable one is "Unused State Variables".

tests/contract-playground/src/StateVariablesManipulation.sol
```diff
@@ -0,0 +1,365 @@
...
+contract NoStateVarManipulationExample {
+    // Simple state variables
+    uint256 public simpleUint;
+    int256 public simpleInt;
+    bool public simpleBool;
+    address public simpleAddress;
+    address payable public simplePayableAddress;
+    string public simpleString;
+    bytes public simpleBytes;
+    IERC20 public h = IERC20(address(3)); // BAD - but not caught. False negative.
```
alexroan, line 17: I've added a false negative here.

alexroan (changes requested):
To fix:
1. A false negative
2. Report noise

TilakMaddy:
@alexroan 

1. False negative addressed
2. Noise removed

alexroan (approved):

PR #673, opened 2024-08-19
Detector: Dead code (Very simple logic)

TilakMaddy:
Severity: Low

Logic
* Internal non overriding functions inside of non abstract contracts that have a body (implemented) and are not used.

Caveat: If an internal function is marked override then, it may still be used even if it doesn't have a direct referencedDeclaration pointing to it. Therefore, we omit the ones marked override


reports/ccip-functions-report.md
TilakMaddy, line 1: These are caught because the only place they are used is in the CCIP `tests/` (and because it's out of scope aderyn sees it as unused)

They are invoked via `super._isTransmitter()` by subclassing this contract.

@alexroan  What do you think we should do here?

TilakMaddy: tldr; functions are not actually used, but they are used in the tests for ccip

alexroan: This happens because `tests/` is excluded. This actually catches bad testing practice, because functions in production contracts should not be made internal purely for the purpose of writing tests. The function should be private, and the developer should find a better way of writing tests. 👍

alexroan (approved):

PR #674, opened 2024-08-19
Detector: Void constructor (ALT. Solution)

TilakMaddy:
Alternative solution to Void Constructor detector by devtooligan

Severity: Low

Logic:

Gather all the invocations of base constructors (by looking at modifiers on constructors of derived contracts)
For each, inspect the contract and see if there is a constructor defined. If there isn't, capture the invocation.


tests/contract-playground/src/VoidConstructor.sol
```diff
@@ -0,0 +1,36 @@
...
+    VoidConstructor() 
+    FilledConstructor() 
+    {
+        
+    }
+}
+
+// BAD invocation (1 instance)
+// NOTE - BUT WE CANNOT CATCH IT (because AST shows no difference between this and the below contract)
+contract C is VoidConstructor() {
```
alexroan, line 25: These bottom two appear exactly the same in the AST? Hmmm That sucks

TilakMaddy: The reason i didn't include that test is because, if you are going to pass in parameters, then that couldve only compiled if the constructor was defined

TilakMaddy: Default constructors are the only ones that can be invoked without defining them .

alexroan (approved):

PR #622, opened 2024-07-28
Detector: State variables not logged events

kree-dotcom:
Added detector for assignment to state variables not recorded by emitted event in the same function. 

Known issues: detector will detect simplistic event emissions only i.e. if state variable x = a then it will detect event emissions which have the argument x or a. If x = a + b  then only event emissions with x will be accepted, if the event records a + b this will cause a false positive.

PR #675, opened 2024-08-19
Detector: Function mutates state variables but does not emit events (ALT Solution)

TilakMaddy:
Alternative solution to #622  

Includes #672 

Severity: low

Qualification: All solc versions

Logic:
Investigate every external/public implemented function.  For each, look for 2 things

```python
Cond A = has any state variable been changed ?
Cond B = has any event been emitted ?


if CondA == YES and CondB == NO :
      capture(self, context, function)
```
 

Thoughts:
Hmm looks like there are a LOT of functions  that change state values but emit 0 events. ... Do we need to rethink our heuristics to reduce that number ??

No associated ADE-Ticket cc: @alexroan


aderyn_core/src/detect/low/state_variable_changes_no_event.rs
TilakMaddy, line 1: See this file

alexroan:
@TilakMaddy please mark as draft for now 🙏

TilakMaddy:
Closing in favor of https://github.com/Cyfrin/aderyn/pull/714/

PR #681, opened 2024-08-19
Fix: Don't report issues with no instances (after aderyn ignore was introduced)

TilakMaddy:
After running through ignore lines, if there are 0 issues, then don't include it in report


aderyn_core/src/lib.rs
```diff
@@ -168,6 +168,11 @@ where
                         .is_some_and(|lines_to_ignore| lines_to_ignore.contains(&instance.1))
                 })
                 .collect();
+
+            if issue.instances.is_empty() {
```
alexroan, line 172: Test for this?

alexroan (changes requested):

TilakMaddy:
Waiting for #704  to get in, so I can write higher quality tests

reports/empty_report.md
TilakMaddy, line 1: Here !

aderyn_core/src/lib.rs
TilakMaddy: ✅ DONE

alexroan (approved):

PR #691, opened 2024-08-21
Fix: nSloc - Correct undercounting of lines + Make filter more robust

### LINKED ISSUE #690: Wrong nSLOC calculation
Running aderyn on this repository reports wrong number of lines

https://github.com/Cyfrin/2024-08-fjord/tree/main

TilakMaddy:
Fix #690

@alexroan , could you also let me know if `cli/reportgen.sh` gives you consistent results on this one ? 

on a separate note (not related to nsloc)
I have discovered some strange things about solc which we didn't account for and I made the fix here
that should help in templegold report giving us more  consistent report.


aderyn_core/src/fscloc/cloc.rs
```diff
@@ -72,6 +80,8 @@ pub fn get_stats(r_content: &str, skip_cloc: bool) -> Stats {
             let prev_group = token_insight_groups.last_mut().unwrap();
 
             if insight.start_line == prev_group.end_line
+                && insight.code_lines.actual_first_line == 0
+                && prev_group.last_token_insight_has_code_in_its_last_line()
```
TilakMaddy, line 84: This condition was missing

aderyn_core/src/fscloc/engine.rs
```diff
@@ -12,16 +12,17 @@ pub fn count_lines_of_code_and_collect_line_numbers_to_ignore(
...
         Box::new(move |res| {
             if let Ok(target) = res {
-                let target_path = target.path().to_str();
                 if target.file_type().unwrap().is_file() {
                     // dbg!(target_path.unwrap());
+                    let remaining = target.path().strip_prefix(&src).unwrap();
                     if src_filepaths
                         .iter()
-                        .any(|fp| target_path.map_or(false, |path| path.contains(fp)))
+                        .any(|fp| &remaining.to_string_lossy().to_string() == fp)
```
TilakMaddy, line 25: Corrected overzealous matching which was the reason for non-deterministic cloc

Inside the `docs/` folder you will see a pattern like `src/FjordToken.sol/something.something`  

Which passed for `src/FjordToken.sol/` and was the reason for error.


That is been fixed now

TilakMaddy: Heck, I made it even better. Instead of filtering out the files of interest, we only feed in the files of interest to the WalkBuilder. That makes it a lot faster because we avoid unnecessary scheduling of threads

aderyn_core/src/fscloc/insight.rs
```diff
@@ -34,21 +35,24 @@ impl From<&TokenDescriptor> for TokenInsight {
...
-        // && insight.code_lines.actual_first_line != -1
-        //{
-        // print!(
-        //     "{:?}, {}-{}\n",
-        //     value.token_type.clone(),
-        //     value.start_line,
-        //     value.end_line
-        // );
-        // println!("{}", value.content.clone());
+        //     && insight.start_line == 3
```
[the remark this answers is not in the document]
TilakMaddy: The  statements helped me a lot with intuition on how I had organized these things and that's why I was able to solve this problem. Is there anything else we can do to avoid removing it from the codebase? 

Surely I can bookmark this commit to see these comments along with where they were .. but just asking for alternative advice ..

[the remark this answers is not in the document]
alexroan: We can leave in here if you feel it's really necessary.

Do these lines provide context that debugging cannot?
Is there context we can add as comments in the function description docblock?

TilakMaddy:
Merged already in #705

PR #694, opened 2024-08-23
Detector: Unchecked low level calls

TilakMaddy:
Severity: High

Logic: below in the comments


tests/contract-playground/src/UncheckedCalls.sol
```diff
@@ -0,0 +1,36 @@
...
+        target.call(abi.encodeWithSignature("someFunction(uint256)", 123));
+
+        target.delegatecall(abi.encodeWithSignature("someOtherFunction(address)", msg.sender));
+
+        target.staticcall(abi.encodeWithSignature("aViewFunction()"));
+    }
+
+    // Copied from Slither
+    function my_func(address payable dst) external payable{
+        dst.call.value(100)("");
```
TilakMaddy, line 34: For any of these calls, if the value is ignored then it's an issue

"call" "delegatecall" "staticcall"

PR #702, opened 2024-08-24
Maintenance: `cargo update`

TilakMaddy:


TilakMaddy:
NOPE ! It fails on my system 😕 ... CI seems to accept it

PR #703, opened 2024-08-24
Fix: Send Ether No Checks (Change satisfaction condition from `msg.sender` to  any address)

TilakMaddy:
Detector:
Send ether no checks

Previous logic:
Look for checks on msg.sender

Flaw:
We can't track data yet. So there are false positives when it comes to say, passing `_msgSender()` as an argument to a function that accepts it as a `address from` in the parameters and then has checks on that.

Mitigation
Changed the logic to be satisfied so long as there is a check on any address


tests/contract-playground/src/Trump.sol
```diff
@@ -0,0 +1,351 @@
...
+        if (from != owner() && to != owner()) {
+            taxAmount = amount.mul((_buyCount>_reduceBuyTaxAt)?_finalBuyTax:_initialBuyTax).div(100);
+
+            if (from == uniswapV2Pair && to != address(uniswapV2Router) && ! _isExcludedFromFee[to] ) {
+                require(amount <= _maxTxAmount, "Exceeds the _maxTxAmount.");
+                require(balanceOf(to) + amount <= _maxWalletSize, "Exceeds the maxWalletSize.");
+
+                if (firstBlock + 3  > block.number) {
+                    require(!isContract(to));
+                }
```
TilakMaddy, line 241: On a side note, this _feels_ like a code smell but I don't why. What do you think this is doing ? :P

aderyn_core/src/detect/high/send_ether_no_checks.rs
```diff
@@ -24,11 +24,11 @@ pub struct SendEtherNoChecksDetector {
...
-            let mut tracker = MsgSenderAndCallWithValueTracker::default();
+            let mut tracker = AddressChecksAndCallWithValueTracker::default();
             let callgraph = CallGraph::new(context, &[&(func.into())], CallGraphDirection::Inward)?;
             callgraph.accept(context, &mut tracker)?;
 
-            if tracker.sends_native_eth && !tracker.has_msg_sender_checks {
+            if tracker.sends_native_eth && !tracker.has_binary_checks_on_some_address {
                 capture!(self, context, func);
             }
         }
```
alexroan, line 34: @TilakMaddy can you please rewrite the title and description? They don't seem quite right.

alexroan (changes requested):

tests/contract-playground/src/Trump.sol
TilakMaddy: Should we write a detector that checks for comparison operator used on `block.number` 🤔

reports/report.md
```diff
@@ -1599,11 +1606,11 @@ The patterns `if (… || true)` and `if (.. && false)` will always evaluate to t
 
 
 
-## H-22: Sending native Eth is not protected from these functions.
+## H-22: Functions send eth away from contract but performs no checks on any address.
 
-Introduce checks for `msg.sender` in the function
+Consider introducing checks for `msg.sender` to ensure the recipient of the money is as intended.
```
TilakMaddy, line 1612: Changed !

PR #704, opened 2024-08-24
Features: Aderyn Ignore Custom detectors

TilakMaddy:


tests/contract-playground/src/AderynIgnoreCustomDetectors.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.20;
```
alexroan, line 2:
```suggestion
pragma solidity 0.8.19;
```

alexroan (approved):
Bro this is excellent!

One small suggestion change and we're good to go.

tests/contract-playground/src/AderynIgnoreCustomDetectors.sol
TilakMaddy: Any reason for this ? I'm missing context ..

TilakMaddy: ✅

PR #705, opened 2024-08-26
Feature: Set default source to `contracts/` for hardhat projects

TilakMaddy:
Set meaningful defaults. I find that more often than not, for hardhat prjects, contracts stay inside `contracts/` directory.
So let that be the default, unless specified otherwise

No assoc. ticket for this (bit adhocy)


TilakMaddy:
Sorry @alexroan  forgot to mention in PR description, this contains #691

PR #710, opened 2024-09-04
LSP integration

TilakMaddy:
Feature: `LSP Server` to work with https://github.com/Cyfrin/vscode-aderyn/ 

Command: `aderyn --lsp --stdout`

Other Changes:
`src_char` now refers characters index starting at the corresponding `line_no`.

How to run?

Step 1: Checkout this branch.
Step 2: Follow steps in extension readme 
Step 3: After pressing F5, open any project and start developing smart contracts :) 

Notes:

1.  LSP server will self-launch
2.  If it fails anytime during your development, press cmd + shift + p to `Restart Aderyn Server`
3.  You'll see patterns that are like `if not lsp_mode { println!(..) }` This is because it's important not to print out random stuff in lsp mode. If you do, client will issue a SHUTDOWN signal and you are toast! Every byte on the stdout should follow lsp spec.

PS:
Also you'll find lightning fast `cargo build` :)


TilakMaddy:
@alexroan Although everything works well, `cli/reportgen.sh` times out and I get unix_wait_error . That goes away when you change the script to run the commands sequentially. (But this is too slow) 😢

PR #714, opened 2024-09-10
State variables manipulated but no event emitted

TilakMaddy:
Same as #675  but cleaner PR


reports/report.md
```diff
@@ -6611,3 +6619,752 @@ State variables that are not updated following deployment should be declared con
...
+- Found in src/Counter.sol [Line: 28](../tests/contract-playground/src/Counter.sol#L28)
+
+	```solidity
+	    function incrementByTwoConstant() external {
+	```
+
+- Found in src/DeadCode.sol [Line: 7](../tests/contract-playground/src/DeadCode.sol#L7)
+
+	```solidity
+	    constructor(uint256 _value) {
```
alexroan, line 6740: We should omit constructors from this detector @TilakMaddy .

alexroan (changes requested):
Just a single change here to omit constructors from this detector!

reports/report.md
TilakMaddy: Done!

alexroan (dismissed):

PR #718, opened 2024-09-13
(Enhancement) Add AsciiArt Install Banner

xyizko:
# Description 

1. The installation of the tool doesnt have a prominent eye catching indicator of what it is. To remedy that an ascii art banner has been added. 
2. The aderyn logo has been converted to ascii art and  is of minimal size such that fidelity is not completely lost 

# AI Usage 

1. Determining the optimal position of placing the banner function so that they are registered during execution.


TilakMaddy:
Hmm ... We plan on having more tools installed in the future on cyfrinup (not just aderyn)

alexroan (approved):
Lovely addition! Thank you @xyizko

PR #720, opened 2024-09-14
imp: Added `deny.toml` for managing licenses and rust crate security from advisories 

TilakMaddy:


alexroan:
Note to self: Leave this til after the detector PRs are merged.

TilakMaddy:
* Removed License checks (headache)

* Kept Crate Bans checks (For now, we add openssl so we make sure release never fails)
* Kept Crate sources checks (Restricted external git sources to aletheia and foundry-config)

PR #722, opened 2024-09-14
Feat: aderyn init

TilakMaddy:
initialize aderyn.toml file


aderyn/src/main.rs
```diff
@@ -109,6 +113,16 @@ fn main() {
...
+    if cmd_args.root == "init" {
+        create_aderyn_toml_file_at(".".to_string());
+        return;
+    }
+
+    if cmd_args.init {
+        create_aderyn_toml_file_at(cmd_args.root);
+        return;
+    }
+
```
TilakMaddy, line 125: I want to support two cases

* aderyn init 

(usually the first argument refers to the project root.  but if it's `init` then we treat it as a special case)

* aderyn optional/path/to/root --init

(expected way to initialize)

Doing this allows us to show users there's an option to init. in the help section. Otherwise if we only have the first case, there's no way to show that

TilakMaddy: TRY THIS

* aderyn init <optional_path_here>

alexroan: I'll wait on this being implemented ^

alexroan (approved):

aderyn/src/main.rs
TilakMaddy: Sorry @alexroan I wasn't done yet, but I'll create an issue to get this done

PR #723, opened 2024-09-14
Fix: Brings support for skipped `unchecked` solidity blocks

TilakMaddy:


reports/templegold-report.md
```diff
@@ -281,15 +281,93 @@ Consider protecting the initializer functions with modifiers.
...
+
+- Found in contracts/util/ABDKMath64x64.sol [Line: 121](../tests/2024-07-templegold/protocol/contracts/util/ABDKMath64x64.sol#L121)
+
+	```solidity
+	      return int128 (result);
+	```
+
+- Found in contracts/util/ABDKMath64x64.sol [Line: 136](../tests/2024-07-templegold/protocol/contracts/util/ABDKMath64x64.sol#L136)
+
+	```solidity
```
TilakMaddy, line 301: These are inside `unchecked` blocks . Weren't caught before, they are now !

alexroan: These are actually false positives, but we should tackle them in a separate issue/pr

alexroan (approved):

PR #724, opened 2024-09-14
Fix: Treat `Expression Statement` as a first class Node

TilakMaddy:
* ExpressionStatement now properly included in the AST
* EventCall was deserialized as an expression, now we explicitly deserialize it as a function call .... (What do you think? Can events be non function call expressions ?)
* UncheckedBlock included in generate capturable methods
* stmt_node! for distinguishing them from normal ast_node!


alexroan (approved):

alexroan:
> (What do you think? Can events be non function call expressions ?)

Based on this grep search in the solidity repo, the references seems to indicate that EventCalls are always FunctionCalls

PR #730, opened 2024-09-17
Skip serializing `src_char2` in JSON report

TilakMaddy:
* Skip serializing `src_char2` in JSON report
(Because it's only used for LSP)

* Get rid of `--skip-build` flag because we no longer rely on `forge build` :)


aderyn_driver/benches/detectors_benchmarks.rs
```diff
@@ -34,7 +34,6 @@ fn bench_aderyn_on_contract_playground(c: &mut Criterion) {
                 root: root_path.to_string_lossy().to_string(),
                 output: String::from("aderyn-report-for-bench.md"),
                 no_snippets: false,
-                skip_build: true,
```
alexroan, line 37: You missed one here @TilakMaddy , added myself 👍

PR #732, opened 2024-09-17
Low Detector: State variable could be marked immutable

TilakMaddy:
Severity: Low

Qualifications: 0.6.5 and above (immutable was introduced)

Logic:
If a state variable has been manipulated how many ever times only in the constructor but not through other entrypoints then it's a candidate for being marked immutable.


tests/contract-playground/src/StateVariableCouldBeDeclaredImmutable.sol
```diff
@@ -0,0 +1,46 @@
...
+    address public h = address(3); // This is a candidate for constant
+
+    uint256 private immutable x;
+
+    constructor() {
+        x = 103; // aderyn-ignore
+        immutableValue = 50; // aderyn-ignore
+        immutableValue *= 3; // aderyn-ignore
+        variableValue = 130; // aderyn-ignore
+        changeSeeminglyImmutableValue();
```
TilakMaddy, line 20: Coping hard lol .. with the `define and use constant`

TilakMaddy:
Woah ... `cli/reportgen.sh` is giving me super inconsistent results

TilakMaddy:
Closing because I'm getting inconsistent results on running cli/reportgen

PR #733, opened 2024-09-17
Fix: visit unchecked block

### LINKED ISSUE #727: Unsafe Casting - Detector should understand `require` and similar checks on unsafe casts
**Detector Title**
Unsafe Casting

**To Reproduce**
See reports/report.md

**Expected behavior**
If an unsafe downcast is first checked with a conditional on acceptable bounds, the instance should not be captured

**Additional context**
[More context](https://github.com/Cyfrin/aderyn/pull/723#discussion_r1761477664)

TilakMaddy:
Forgot to add this in the workspace visitor file.

 Fixes #727 

Actually, we do have check for binary operation (comparison checks on the variable) before reporting unsafe conversion

The problem was I forgot to include the visitor function in workspace file.

I have added tests to prove my case


tests/contract-playground/src/Casting.sol
```diff
@@ -141,4 +148,20 @@ contract Casting {
...
+
+    function from128x128_checked(int256 x) internal pure returns (int128) {
+        unchecked {
+            int256 result = x >> 64;
+            // GOOD
+            require(result > 1 && result < 100, "BAD range given");
+            return int128(result);
+        }
+    }
+}
```
TilakMaddy, line 167: Here you can see we only catch the BAD ones

reports/templegold-report.md
```diff
@@ -282,93 +282,15 @@ Consider protecting the initializer functions with modifiers.
...
-	      return int128 (result);
-	```
-
-- Found in contracts/util/ABDKMath64x64.sol [Line: 151](../tests/2024-07-templegold/protocol/contracts/util/ABDKMath64x64.sol#L151)
-
-	```solidity
-	      return int128 (result);
-	```
-
-- Found in contracts/util/ABDKMath64x64.sol [Line: 232](../tests/2024-07-templegold/protocol/contracts/util/ABDKMath64x64.sol#L232)
```
TilakMaddy, line 312: These ones undergo binary checks. So they won't be caught

aderyn_core/src/visitor/workspace_visitor.rs
```diff
@@ -151,6 +151,7 @@ impl ASTConstVisitor for WorkspaceContext {
         visit_structured_documentation | StructuredDocumentation => structured_documentations_context |,
         visit_tuple_expression | TupleExpression => tuple_expressions_context |,
         visit_unary_operation | UnaryOperation => unary_operations_context |,
+        visit_unchecked_block | UncheckedBlock => unchecked_blocks_context |,
         visit_user_defined_value_type_definition | UserDefinedValueTypeDefinition => user_defined_value_type_definitions_context |,
```
TilakMaddy, line 155: CULPRIT !!

reports/report.md
```diff
@@ -460,567 +460,573 @@ Consider protecting the initializer functions with modifiers.
...
 	```solidity
 	        bytes1Value = bytes1(ae);
 	```
 
+- Found in src/Casting.sol [Line: 155](../tests/contract-playground/src/Casting.sol#L155)
+
+	```solidity
+	            return int128(result);
+	```
+
```
TilakMaddy, line 1029: Caught HERE

reports/templegold-report.md
alexroan: Nice!

alexroan (approved):
Nice find!

PR #734, opened 2024-09-18
Low Detector: State variable could be marked immutable + Inter Context Merge Strategy 

TilakMaddy:
Severity: Low

Qualifications: 0.6.5 and above (immutable keyword was introduced)

Logic:
If a state variable has been manipulated how many ever times only in the constructor but not through other entrypoints then it's a candidate for being marked immutable.


aderyn_core/src/detect/low/function_pointer_in_constructor.rs
```diff
@@ -99,6 +99,22 @@ mod func_compilation_solc_pragma_helper {
             }
             false
         }
+        pub fn compiles_for_solc_below_0_6_5(&self, context: &WorkspaceContext) -> bool {
+            if let Some(source_unit) = self.closest_ancestor_of_type(context, NodeType::SourceUnit)
+            {
+                let pragma_directives = ExtractPragmaDirectives::from(source_unit).extracted;
+
+                if let Some(pragma_directive) = pragma_directives.first() {
```
TilakMaddy, line 107: We'll bring these helper functions into a helpers module in a separate PR so it's visible

tests/contract-playground/src/inheritance/ExtendedInheritance.sol
```diff
@@ -15,6 +15,7 @@ contract ExtendedInheritance is InheritanceBase {
         for (uint256 i = 0; i < 3; i++) {
             target.delegatecall(abi.encodeWithSignature("doSomething(uint256)", i));
         }
+        s_baseAddress = target;
```
alexroan, line 18: I've added this to test the case where inheriting contracts use the state variables outside of the constructor. This should not be caught, but it is.

reports/report.md
```diff
@@ -7226,7 +7252,210 @@ State variable changes in this function but no event is emitted.
...
+- Found in src/eth2/DepositContract.sol [Line: 66](../tests/contract-playground/src/eth2/DepositContract.sol#L66)
+
+	```solidity
+	    bytes32[DEPOSIT_CONTRACT_TREE_DEPTH] zero_hashes;
+	```
+
+- Found in src/inheritance/InheritanceBase.sol [Line: 9](../tests/contract-playground/src/inheritance/InheritanceBase.sol#L9)
+
+	```solidity
+	    address public s_baseAddress;
```
alexroan, line 7433: This should not be caught. If this were made immutable, the child contract would not compile.

alexroan (changes requested):
This catches some false-positives @TilakMaddy .
Can you please account for child contracts that reference and set state variables?

reports/report.md
TilakMaddy: Good catch! I'll work on resolving this

aderyn_core/src/detect/detector.rs
```diff
@@ -489,4 +503,8 @@ pub trait IssueDetector: Send + Sync + 'static {
     fn hints(&self) -> BTreeMap<(String, usize, String), String> {
         BTreeMap::new()
     }
+
+    fn merge_conflict_resolution_strategy(&self) -> MergeConflictResolutionStrategy {
```
alexroan, line 507: Could you explain in more detail why this is needed @TilakMaddy ?

Is this something we'll need to add to other detectors?

TilakMaddy: Yeah, so this depends on the nature of the detector. By default all detectors implement the Union Strategy however, this one should implement Intersection Strategy.

The idea behind this is that, say you have 2 files:
File 1: pragma solidity 0.8.20
(Think ChildContract)

File 2: pragma solidity ^0.8.0
(Think parent contract)

And code in file1 imports some stuff from file2, we will have the following contexts:

Context 1:
File1, File2

Context2:
File2

Notice that File2 appears in both contexts (we already knew this from ICF so not going in detail here)

The important part is that when a detector finds that some variable could be marked immutable just by looking at File2 in Context2 that doesn't make it safe to conclude that the variable could in fact be marked immutable.
The reason is that when the same detector looks at File2 from Context1, it will see that nothing can be marked immutable. 

Therefore, for the same file, for the same detector we have 2 different answers from 2 different contexts.
This is a case of merge conflict. So far, what we have done is reported the union of all issue instances. (Which is still the default strategy).

However here in this detector, we can only safely conclude that a variable could be marked immutable if for a given file, all the contexts that involve the said file, report the same. (We must take intersection here)
Therefore due to the nature of this detector it is given Intersection resolution strategy

@alexroan

TilakMaddy:
FOLLOW UP:

Use Intersection strategy for unused imports, unused errors as well

aderyn_core/src/detect/detector.rs
```diff
@@ -442,6 +447,15 @@ pub enum IssueSeverity {
     High,
 }
 
+/// When different contexts report different number of instances for the same file, this is a merge
+/// conflict and we need a resolution strategy.
+pub enum MergeConflictResolutionStrategy {
```
alexroan, line 452: @TilakMaddy could you please remove this as discussed, and use only intersection?

alexroan (dismissed):

aderyn_core/src/detect/detector.rs
TilakMaddy: Until then, I'l draft it ... Let me know if you disagree

TilakMaddy: Actually I'll do what you said - remove this strategy enum.

When you come back later, we'll see

TilakMaddy:
@alexroan Although I have made the changes you have requested but I might have to click on "dismiss review" here for a second here .. so I can merge :P Hope  you don't mind

PR #735, opened 2024-09-18
Low Detector: Libraries and contracts should not live in same file

TilakMaddy:
Severity: Low

Qualifications: All versions on solc

Logic:
If library contract is seen in a file that has other non-library contracts, we capture the library contract and suggest that it be moved to a different file.

Reason: Library contracts generally need floating pragmas to support a wide variety of contracts. If they are in a file same as contract, then this distinction is hard to support

No assoc. ADE

FOLLOW UP HERE 🙏  #740


TilakMaddy:
We need flexibility to organize solidity contracts and libraries the way we want (also helps for security scanning and verification)

PR #736, opened 2024-09-18
Low Detector: Event emitting a public constant in non constructor functions

TilakMaddy:
Severity: Low

Qualifications: All solc version

Logic:
Event emitting a public constant from a  non constructor function 

It's been on my mind for a while.. I thought I'll try this out 

No assoc. ADE


alexroan:
There will be some cases where this information is useful for offchain indexing IMO. Closing.

PR #738, opened 2024-09-18
Low Detector: Solidity 0.8.23 introduces MCOPY opcode

TilakMaddy:
Severity: Low

Qualifications: If contract's pragma directives allow it to compile with solc 0.8.23 version or above, then warn users that not all chains support the potentially introduced MCOPY opcode


TilakMaddy:
Stale PR

We don't have support for checking solc_version before making this claim. Also the PR is stale

PR #755, opened 2024-10-04
standardize rust formatting with `rustfmt.toml` 

TilakMaddy:
Reference:

1. https://github.com/foundry-rs/foundry/blob/master/.git-blame-ignore-revs
2. https://github.com/foundry-rs/foundry/blob/master/rustfmt.toml

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,11 @@
+reorder_imports = true
+imports_granularity = "Crate"
+use_small_heuristics = "Max"
+comment_width = 100
+wrap_comments = true
+binop_separator = "Front"
+trailing_comma = "Vertical"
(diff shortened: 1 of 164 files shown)


rustfmt.toml
```diff
@@ -0,0 +1,11 @@
...
+imports_granularity = "Crate"
+use_small_heuristics = "Max"
+comment_width = 100
+wrap_comments = true
+binop_separator = "Front"
+trailing_comma = "Vertical"
+trailing_semicolon = false
+use_field_init_shorthand = true
+format_code_in_doc_comments = true
+doc_comment_code_block_width = 100
```
TilakMaddy, line 11: This guy does the magic !

Issue #778, opened 2024-11-11
`cyrfinup/install` doesn't work with non-interactive zsh shells


0ximmeas:
**Describe the bug**
`.zshrc` is [only sourced](https://unix.stackexchange.com/a/71258) in interactive shells. This makes it difficult to use the install script and `cyfrinup` in non-interactive shells, for example, Dockerfiles. `.zshenv` is sourced in all shells hence a better file to use.

**To Reproduce**
Add the following to a [Dockerfile](https://github.com/0ximmeas/devcontainer/blob/main/.devcontainer/Dockerfile) and run it:
```dockerfile
RUN curl -L https://raw.githubusercontent.com/Cyfrin/aderyn/dev/cyfrinup/install | zsh
RUN cyrfinup
```

**Expected behavior**
For `cyfrinup` to be sourced in non-interactive zsh shells.

**Screenshots**
n/a

**Desktop:**
n/a

**Additional context**
As a reference, [this](https://github.com/foundry-rs/foundry/blob/master/foundryup/install#L25) is the same way that `foundryup` is sourced.

alexroan:
Hey @0ximmeas , this should be fixed in #787 , which has been merged to dev.

When the next release is published, this should be fixed!

PR #779, opened 2024-11-11
fix: make install work with non interactive zsh shells

0ximmeas:
fix for #778, `.zshenv` is sourced in all shells, including non-interactive


0ximmeas:
Manually tested in both interactive and non-interactive zsh shell (and it's the same way as [`foundryup`](https://github.com/foundry-rs/foundry/blob/master/foundryup/install#L25) does it)

alexroan:
Thank you @0ximmeas , resolved in #787 :)

PR #802, opened 2025-02-20
Fix the version for aderyn.toml + Optimize CI by canceling in-progress tasks running concurrently in a group

TilakMaddy:
Blocked due to dependency 

@alexroan  ---> https://github.com/alexroan/2024-05-Sablier/pull/1


TilakMaddy:
@alexroan For now, the following behavior:

If version field is not specified, don't care otherwise make sure it is set to 1.

Later, we can make this field mandatory

PR #810, opened 2025-03-06
Audit mode: Entry Points

alexroan:
* New audit detector: Entry Points
  * All public and external entry points into the contracts in the workspace
* Rename `attack-surface` to `raw-calls` 
* CLI behaviour change
  * `aderyn --audit` - runs all audit mode detectors
  * `aderyn --audit <detector_name>` runs specific audit mode detector. Example:
  * `aderyn --audit entry-points`

### DIFF
--- a/aderyn/src/main.rs
+++ b/aderyn/src/main.rs
@@ -76,8 +76,9 @@ pub struct CommandLineArgs {
     skip_update_check: bool,
 
     /// Run in Auditor mode, which only outputs manual audit helpers
-    #[arg(long)]
-    auditor_mode: bool,
+    /// Optional value can specify a specific auditor detector to run (e.g. "entry-points")
+    #[arg(long, num_args = 0..=1, default_missing_value = "all")]
+    audit: Option<String>,
 }
 
 #[derive(Debug, Subcommand)]
@@ -147,7 +148,7 @@ fn main() {
         skip_cloc: cmd_args.skip_cloc,
         skip_update_check: cmd_args.skip_update_check,
         stdout: cmd_args.stdout,
-        auditor_mode: cmd_args.auditor_mode,
+        audit: cmd_args.audit,
         highs_only: cmd_args.highs_only,
         lsp: cmd_args.lsp,
     };
--- a/aderyn_core/src/audit/entry_points.rs
+++ b/aderyn_core/src/audit/entry_points.rs
@@ -82,12 +82,15 @@ impl AuditorDetector for EntryPointsDetector {
 
 #[cfg(test)]
 mod entry_points_test {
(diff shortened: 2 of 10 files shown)


aderyn/src/main.rs
```diff
@@ -76,8 +76,9 @@ pub struct CommandLineArgs {
     skip_update_check: bool,
 
     /// Run in Auditor mode, which only outputs manual audit helpers
-    #[arg(long)]
-    auditor_mode: bool,
+    /// Optional value can specify a specific auditor detector to run (e.g. "entry-points")
+    #[arg(long, num_args = 0..=1, default_missing_value = "all")]
+    audit: Option<String>,
```
TilakMaddy, line 81: Does this default value affect the output we get when we run just  `aderyn` ?
What happens when we do a `-o report.json --stdout` ?

aderyn_core/src/audit/entry_points.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+impl AuditorDetector for EntryPointsDetector {
+    fn detect(&mut self, context: &WorkspaceContext) -> Result<bool, Box<dyn Error>> {
+        let functions = get_implemented_external_and_public_functions(context);
+        functions.for_each(|function_definition| {
+            if let ASTNode::ContractDefinition(contract_definition) = context
+                .get_closest_ancestor(function_definition.id, NodeType::ContractDefinition)
+                .unwrap()
+            {
```
TilakMaddy, line 49: Use `if let Ok(..)  = {}`

TilakMaddy: Or `Some()` ...I don't remember which one...but one of those

aderyn/src/main.rs
TilakMaddy, line 81: Should this CLI flag be considered a breaking change? (in other words next release would this require us to bum up minor version in the next release of aderyn cli)?

If yes, then we need to make the change in vscode-extension too . Refer here https://github.com/Cyfrin/vscode-aderyn/blob/main/RELEASE_CHECKLIST.md

TilakMaddy:
You might also want to check linearized base contracts ....

aderyn/src/main.rs
TilakMaddy: Well I don't think so here, because it's `default_missing_value` and not `default_value` : )

Also, no changes to reports. So looks good to me

alexroan: Yes, since it's a change in user-facing behaviour

TilakMaddy:
I think we should draft this PR, IMO we'll focus on this at a later point in the roadmap..

PR #819, opened 2025-03-10
feat: remove remappings from aderyn.toml

TilakMaddy:
To make it less confusing, ask the users to write the remappings in `remappings.txt`. That should be enough.

If remappings were filled (for the minority of users to who have used this feature), we  can display a warning saying...these will not be considered. Please specify in remappings.txt.
(wip)


TilakMaddy:
Done in some other PR

PR #820, opened 2025-03-10
Add env variable support to `aderyn.toml`

TilakMaddy:
* No breaking changes

* Tests included

Motivation:
So far, on command line, there are no issues. Because the env is inherited from the shell into the current process. For vscode extension, we need to explicitly make this inheritance. But we don't know which shell the user has set all the variables. Therefore, this config allows for the agreement to be made explicit


tests/foundry-nft-f23-icm/aderyn.toml
```diff
@@ -0,0 +1,8 @@
+# Aderyn Configuration File
+# Help Aderyn work with more granular control
+
+# DO NOT CHANGE version below. As of now, only 1 is supported
+version = 1
+
+[env]
+FOUNDRY_PROFILE = "icm"
```
TilakMaddy, line 8: This is what the PR allows you to do

TilakMaddy:
@alexroan New foundry project added as integration test here. You might want to update your git submodule :)

PR #828, opened 2025-03-11
Rename and Re-describe detectors

### LINKED ISSUE #818: Rename detectors
Example:

useless-public-function --> overexposed-public-function 

Because the former sounds like you can get rid of the function which is not what we want to imply.

alexroan:
* Fixes #818 
* Remove Indexed Events detector (too noisy - no developer gain)


reports/report.md
```diff
@@ -233,15 +232,14 @@ This report was generated by [Aderyn](https://github.com/Cyfrin/aderyn), a stati
 | Category | No. of Issues |
 | --- | --- |
 | High | 38 |
-| Low | 48 |
+| Low | 47 |
```
alexroan, line 235: Removed indexed events detector ^

aderyn_core/src/detect/detector.rs
```diff
@@ -115,82 +114,81 @@ pub fn get_all_detectors_names() -> Vec<String> {
...
-    ContractWithTodos,
+    Todo,
     InconsistentTypeNames,
     UnprotectedInitializer,
-    RevertsAndRequiresInLoops,
+    RequireRevertInLoop,
     DivisionBeforeMultiplication,
-    UnsafeCast,
+    UnsafeCasting,
     EnumerableLoopRemoval,
```
TilakMaddy, line 160: Do we want to specify that it's only for integers ?

TilakMaddy: Unsafe Cast I mean ..

TilakMaddy (approved):
LGTM

aderyn_core/src/detect/detector.rs
alexroan: Na, that's specified in the description. I prefer these names to be shorter tbh.

PR #845, opened 2025-03-16
Feat/remove remappings support

TilakMaddy:
`remappings` ❌ 
`remappings.txt` ✅


TilakMaddy:
@alexroan I need to know what you think of this.

Goal: **To keep everything simple**

remappings.txt seems to be the standard.

TilakMaddy:
This will be part of same Pr as integration with ICF 2.0

PR #850, opened 2025-03-19
edit: Readme -  Addition of cli gif demo 

xyizko:
# What ? 

1. A simple compressed gif demo of the cli app has been added to the readme.md
2. A demo contract with a `re-entrancy vulnerability` was scanned
3. The gif is located in the same place with other image assets


TilakMaddy:
Thanks but I'm not keen on merging this. Mainly because I'd like to see attention go to the vscode extension link that we have in the readme. 

Plus, there are sample reports [here](https://github.com/Cyfrin/aderyn/tree/dev/reports) so it should be easy for people to see what the reports look like.

PR #859, opened 2025-03-22
chore: dedup logic in makefile

TilakMaddy:


reports/report.json
TilakMaddy, line 1: https://github.com/Cyfrin/aderyn/pull/858#discussion_r2008799426

Rectified here

PR #868, opened 2025-03-23
ci: performance improvement

TilakMaddy:


TilakMaddy:
1 minute for integration tests now ? 🔥 
cc: @alexroan

PR #871, opened 2025-03-25
Fix/fp ignore

TilakMaddy:


tests/contract-playground/src/AderynIgnoreCustomDetectors.sol
```diff
@@ -39,4 +39,10 @@ contract AderynCustomIgnore {
     function f7() public {
 
     }
+
+    // This will be not reported by either of the detectors
+    // aderyn-fp-next-line(unused-public-function) aderyn-ignore-next-line(empty-block)
+    function f8() public {
+
```
TilakMaddy, line 46: This was not possible before, now it is.

Issue #836, opened 2025-03-13
OpenZeppelin UUPSUpgradeable reinitializer function triggers Unprotected Initializer


studiov-dev:
**Detector Title**
H - Unprotected Initializer

**To Reproduce**
1. Clone repo: `https://github.com/Studio-V-Tech/neulock-onchain.git`
2. Run `npm i`
3. Run aderyn with this: `aderyn -x mocks,harnesses,old`
4. Scroll down to [H-1: Unprotected initializer]
5. See error

**Screenshots**
N/A

**Expected behavior**
Issue should not be detected, because function uses the `reinitializer(2)` modifier from OpenZeppelin UUPSUpgradeable lib.

**Additional context**
The tests in `npm run test test/Upgrades.ts` pass, including `Reverts on running reinitialize function on V2 again`, which checks that the `reinitializer()` modifier works as expected.

PR #874, opened 2025-03-27
fix: Add support for OpenZeppelin UUPSUpgradeable reinitializer function

sumitvekariya:
## Description
This PR fixes issue #836, where Aderyn was generating false positives for the "Unprotected Initializer" detector when analyzing contracts that use OpenZeppelin's UUPSUpgradeable `reinitializer(n)` modifier.

The detector is now enhanced to recognize both `initializer` and `reinitializer` modifiers as adequate protection for initialization functions, thus reducing false positives.

## Changes
- Modified the `UnprotectedInitializerDetector` to check for modifiers whose names start with "reinitializer", in addition to the existing check for the "initializer" modifier.
- Added a test contract (`ReinitializerTest.sol`) with both protected and unprotected initializer functions to validate the changes.
- Added a test case to ensure the fix works correctly.

## Testing
The changes have been tested with:
- A new test case specifically for the `reinitializer` modifier
- The existing test case for the `initializer` modifier
- The full test suite to ensure no regressions

## Related Issue
Fixes: #836


tests/contract-playground/src/ReinitializerTest.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.19;
+
+// Mock OpenZeppelin UUPSUpgradeable contract with reinitializer
+abstract contract Initializable {
+    /**
```
TilakMaddy, line 7: Is it possible to use real contract as shown [here](https://github.com/Cyfrin/aderyn/blob/dev/tests/contract-playground/src/UnprotectedInitialize.sol) ?

Also see if you can create the ReinitializerContract in the same file above...

aderyn_core/src/detect/high/unprotected_initializer.rs
```diff
@@ -90,12 +90,12 @@ impl CallGraphVisitor for UnprotectedInitializationTracker {
         for inv in modifier_invocations {
             match inv.modifier_name {
                 crate::ast::IdentifierOrIdentifierPath::Identifier(n) => {
-                    if n.name == "initializer" {
+                    if n.name == "initializer" || n.name.starts_with("reinitializer") {
                         self.has_initializer_modifier = true;
```
TilakMaddy, line 94: Nit: Can you do an exact match instead of starts_with ?

TilakMaddy:
LGTM

PR #898, opened 2025-04-07
Chore/cleanup 2

TilakMaddy:


aderyn_driver/src/runner.rs
```diff
@@ -22,18 +22,14 @@ pub fn run_detector_mode<T>(
...
 {
-    let mut ignore_lines = HashMap::new();
-    for context in contexts {
-        ignore_lines.extend(context.ignore_lines_stats.clone());
-    }
     println!("Running {} detectors", detectors.len());
 
     let detectors_used =
         &detectors.iter().map(|d| (d.name(), d.severity().to_string())).collect::<Vec<_>>();
-    println!("Detectors run, processing found issues");
```
TilakMaddy, line 33: This line has to run AFTER the detectors are done running. So this PR fixes that

PR #918, opened 2025-04-11
feat(`aderyn init`): Guess the root by looking one directory deep and auto add that to init file

TilakMaddy:


TilakMaddy:
TODO:
write CI tests

PR #922, opened 2025-04-12
Master merge v0.5.7

TilakMaddy:


TilakMaddy:
Ahhh now I see, merge conflict checker doesn't look inside `.github` /

the tests failing was not supposed to be there. I deleted that in dev and even the production binary works fine.

cool!

PR #933, opened 2025-04-22
Update github actions versions due to deprecation

PixelPil0t1:
Artifact Actions v3 will be deprecated by January 30, 2025. GitHub has started temporarily failing workflows that haven't been updated ahead of this deadline. This PR updates the action versions accordingly, following the [official migration guide](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md).

### DIFF
--- a/.github/workflows/web.yml
+++ b/.github/workflows/web.yml
@@ -53,7 +53,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       # Setup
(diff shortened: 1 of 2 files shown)


TilakMaddy:
Sadly, we don't run either of the workflows 😅 altered in this PR...
but it's okay maybe in the future we'll consider running web.yml

Issue #1020, opened 2025-12-30
Ignore lines are... ignored


PatrickAlphaC:
I have the following report:

```
## L-2: Loop Contains `require`/`revert`

Avoid `require` / `revert` statements in a loop because a single bad item can cause the whole transaction to fail. It's better to forgive on fail and return failed elements post processing of the loop

<details><summary>2 Found Instances</summary>


- Found in src/Agreement.sol [Line: 115](src/Agreement.sol#L115)

	```solidity
	        for (uint256 i = 0; i < caip2ChainIds.length; ++i) {
	```

- Found in src/Agreement.sol [Line: 159](src/Agreement.sol#L159)

	```solidity
	        for (uint256 i = 0; i < chains.length; ++i) {
	```

</details>

```

But if I put an ignore statement on `src/Agreement.sol#L115`, I still get this report out.

# Reproduce

```
git clone https://github.com/Cyfrin/battlechain-safe-harbor
cd battlechain-safe-harbor
git checkout b35e6f13d20eb83fcf38eb18e48ce028a2713094
forge build
aderyn .
```

If you add ignore lines on either of those flagged issues, it will still flag them.

TilakMaddy:
I have checked battlechain with the above change. Can confirm there it is not a bug on Aderyn's side

PR #1021, opened 2026-01-07
Fix ignore regex to support ignore-all comments

developerfred:
## Summary
Fixes issue #1020: Ignore lines are... ignored

The regex for parsing aderyn-ignore comments was requiring '(' even for ignore-all comments without detector specification. This has been fixed by making the detector specification optional.

## Changes
- Modified the regex in aderyn_core/src/stats/ignore.rs to make the detector list optional

## Test Results
All tests pass successfully.

```
running 1 test
test parse_comments_to_rightfully_ignore_lines::test_aderyn_ignore_specific_detectors ... ok
```


TilakMaddy:
Hey sorry this issue was meant to be solved by internal members. (As you can see the codebase mentioned is not a public one, so reproducing that is not in scope for public contributors)

